PR #101, opened 2022-06-01
Fix `getAllPools` taking too long

### LINKED ISSUE #100: query.getPools() takes 1.5h on rinkeby
**Problem**

@trizin: Seems like get pools is working but getAllPools is taking too long to finish. I added `tqdm` to range. It takes ~1.5 hours to get all pools.

@trizin: it is scanning all blocks in the network

**TODOs**

Find a way to make it faster.

trizin:
Fixes #100

Changes proposed in this PR:

- Use a while-ish loop to get all pools (1k in a batch), break when there are no pools left.

### DIFF
--- a/util/query.py
+++ b/util/query.py
@@ -2,7 +2,6 @@ import json
 from typing import Dict, List, Tuple
 
 import requests
-import brownie
 from enforce_typing import enforce_types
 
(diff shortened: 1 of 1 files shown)


trentmc (approved):
nice!

PR #104, opened 2022-06-02
Use token address instead of symbol

### LINKED ISSUE #73: Attack: risk of fake volumes & stakes if a fake token with symbol=OCEAN is used
**Describe the bug**

query.py::.query() computes `volumes` with getPoolVolumes() and getDTVolumes(). Their results are keyed by basetoken symbol not basetoken address.

And it may be the same for `stakes` too.

**To fix**

Key `volumes` by basetoken address (not symbol).

If appropriate, key `stakes` by basetoken address.

trizin:
Fixes #73

Changes proposed in this PR:

- Update `getStakes` and `getVolumes` to use token address instead of the symbol.
- [Connect to network before `stakesToUsd` and `volumesToUsd`](https://github.com/oceanprotocol/df-py/pull/104/commits/5b85fe07ec20ca440f4205abd9f00ac63887e689)
- Updated tests

### DIFF
--- a/util/test/test_query.py
+++ b/util/test/test_query.py
@@ -12,7 +12,7 @@ from util.constants import BROWNIE_PROJECT as B
 account0, QUERY_ST = None, 0
 
 CHAINID = networkutil.DEV_CHAINID
-OCEAN_ADDR: str = None
+OCEAN_ADDR: str = ""
 
 
 @pytest.mark.timeout(300)


util/calcrewards.py
```diff
@@ -4,6 +4,8 @@
 from numpy import log10
 
 from util import cleancase
+from util.query import _symbol
```
trentmc, line 7: Given that `_symbol()` is being used outside the original module, please add a unit test for it, and rename to `symbol()`

trentmc (approved):

trentmc (approved):

PR #108, opened 2022-06-03
Feature/claim and stake

alexcos20:
Fixes # .

Changes proposed in this PR:

- 
- 
-


trentmc:
trentmc June 3: 
Alex, why re-implement batch stake here, rather than use the one provided by the router?  (edited) 

alex June 3:

1. you need to hardcode router address somewhere
2. more expensive in gas (you need to transfer Ocean from df-reward -> router -> pool)
3. you need the loop anyway, because you will need to transfer shares back to user

trentmc: 

Alex - great answer, thanks!

trentmc:
> Should we add a swap feature inside the `claimAndStake` function in case the reward token and the pool token are different? Perhaps it would be better to hide the "Claim & re-stake" button for such occurrences.

Doing that is outside the scope of this issue. It's in [df-issues#69](https://github.com/oceanprotocol/df-issues/issues/69). Doing a swap is only one possible approach. And there's other trickiness and complexities; we need to be thoughtful when we solve it. Hence, do _not_ do for _this_ issue.

PR #109, opened 2022-06-03
Check after-before balances

### LINKED ISSUE #106: Unchecked balances
When toping up the contract, please make sure that you are checking the actual balance before and after the transfer (https://github.com/oceanprotocol/df-py/blob/main/contracts/DFRewards.sol#L36-L40)
And balanceBefore+total_value should match balanceAfter

There are numerous exploits with tokens that have feeTransfers

trizin:
Fixes #106 

Changes proposed in this PR:

- Make sure contract balance after - before is equal to `total_value`


trentmc (changes requested):
You should also add a unit test for this

trentmc (approved):

PR #128, opened 2022-06-09
Fix networkutil

idiom-bytes:
Fixes #NA

General fixes towards deployment of prod


trentmc (changes requested):
Removing the f in front of the f string in line 47 will break line 47 functionality

idiom-bytes:
> Um okay

Too many similar moons and chainIds.
Let's make sure they are the right ones.

trentmc (approved):

PR #132, opened 2022-06-10
Fix `getDTVolumes` loop

### LINKED ISSUE #131: `getDTVolumes` loop logic is incorrect
The for loop inside the `getDTVolumes` function should stop once there are no orders left.

trizin:
Fixes #131

Changes proposed in this PR:

- [Fix getDTVolumes loop](https://github.com/oceanprotocol/df-py/commit/6efd1817537f6747308c037d06b0e6107f857d4f)
- [Update getAllPools](https://github.com/oceanprotocol/df-py/commit/a8d8bd8b2322d417f14960fa29d16c94156872c6)
- Use a while loop and break when there are no orders left.

### DIFF
--- a/util/query.py
+++ b/util/query.py
@@ -233,7 +233,7 @@ def getDTVolumes(
             offset,
             chunk_size,
         )
-        offset += 1000
+        offset += chunk_size
         result = submitQuery(query, chainID)
         new_orders = result["data"]["orders"]
         if new_orders == []:


util/query.py
```diff
@@ -232,8 +233,11 @@ def getDTVolumes(
             offset,
             chunk_size,
         )
+        offset += 1000
         result = submitQuery(query, chainID)
         new_orders = result["data"]["orders"]
+        if new_orders == []:
```
trentmc, line 239: This relies on orders to be in every chunk, I think. 

Ie i think it will prematurely terminate if there's a chunk that doesn't have orders.

trentmc: >  we can assume the rest of the chunks will be empty once we get an empty chunk.

I question whether we can.

Let's use the following data as a working example. 
- chunk 0: block 0 - 999: 12 orders
- chunk 1: block 1000 - 1999 - 17 orders
- chunk 2: block 2000 - 2999 - 0 orders
- chunk 3: block 3000 - 3999 - 25 orders
- chunk 4: block 4000 - 4231 - 10 orders (chain length = 4231 blocks)

There are two different possible assumptions about contents of `new_orders` data structure:
- a. "`new_orders` is always length 1000, except for very last one"
- b. "`new_orders` may be length 0, or 1, ..., or 1000, depending on how many actual orders

Your code so far assumes (a). I'm pretty sure it's (b), though I could be wrong! 

If (b), then the example data would stop at chunk 2. Then chunks 3 and 4 would be missed.

Could you dive deeper to establish definitively whether (a) vs (b)? Perhaps write a test where it's got data set up like above.

----

BTW if (b), then of possible boolean tests, approach 3 below is best:

```python
#approach 1
if new_orders == []:
  break

#approach 2
if len(new_orders) != 1000:
  break

#approach 3
if not new_orders:
    break
```

trentmc: > That's right but the block number doesn't change, the range is static (last Thursday - now).

Ah right, this method is within a fixed block range.

But we still have an issue. Within a given block range, if on say offset = 2000, it may have no orders, yet on say offset  3000, there may be orders. The code as-is would miss the orders on offset 3000.

----
Also, small thing: it shouldn't be `offset += 1000`, it should be `offset += chunk_size`.

trentmc (changes requested):

util/query.py
trentmc: > The list is ordered, we should get the next orders in the list. If offset = 2000 is empty we can say there are less than 2000 orders and offset = 3000 would be empty too

OK. That clarifies. Then, good to go I believe!

trentmc (approved):

PR #133, opened 2022-06-10
Include pools with zero volume in poolinfo CSV

### LINKED ISSUE #127: Update dftools/dfpy query to not filter 0 consume pools
Problem:
- If we consume manually to populate pools, next week it might be done
- There might be good data sets with liquidity but no consume, and it will lead to a "search/discovery" problem because they are technically valid, but have no consume

Recommendation:
- When querying subgraph from dftools, include 0 consume pools.
- Sort pools in table by consume (descending) so we have high-consume, or high-stake, or high-rewardFunction pools listed first

trizin:
Fixes #127

Changes proposed in this PR:

- Include pools with zero volume in poolinfo CSV

### DIFF
--- a/util/csvs.py
+++ b/util/csvs.py
@@ -267,7 +267,6 @@ def savePoolinfoCsv(
                 did = oceanutil.calcDID(p.nft_addr, chainID)
                 url = constants.MARKET_ASSET_BASE_URL + did
 
-                stake_amt = 0
                 stake_amt = sum(stakes_at_chain[basetoken][pool_addr].values())
 
                 vol = 0


util/csvs.py
```diff
@@ -258,26 +258,35 @@ def savePoolinfoCsv(
...
-                row += [stake_amt]
 
                 p = pools_by_addr[pool_addr]
-                row += [p.nft_addr, p.DT_addr, p.DT_symbol, p.basetoken_addr]
-
                 did = oceanutil.calcDID(p.nft_addr, chainID)
                 url = constants.MARKET_ASSET_BASE_URL + did
-                row += [did, url]
 
+                stake_amt = 0
```
trentmc, line 270: Don't need this line, since the line after fully defines `stake_amt`

trentmc (approved):
All good! Just make the one-line change before merging. Thanks:)

PR #141, opened 2022-06-14
Issue 137 - Split DFRewards.sol into >1 contracts

### LINKED ISSUE #137: Split DFRewards.sol into >1 contracts, to reconcile security * future_migration
Trent:

Manan had a cool idea for DFRewards.sol, around the question of:

1. how do you have good smart contract security
2. allow for new changes to the smart contract, and have good ux

- If you use EIP1167 proxy contract on DFRewards.sol, we can do upgrades (2) but security is so-so (1).
- If we hardcode DFRewards.sol, we have great security (1) but if we want to upgrade we have to re-deploy (2).
- Solution: Have _two_ contracts. One stores the mapping of who gets what; it can't be upgraded. The other contract has different interface-y stuff, and it can get upgraded.

Thoughts?

----
Roberto:

I like the idea of separating concerns to different contracts.

Keeping a rewards mapping/balance somewhere else seems like a good improvement however as you said, creates more surface area for an attack if the SLA/function access isn't properly tackled.

Personally, I'm more concerned with other functional requirements (a decentralized partner treasury, including Ocean) so we don't have to manage funds manually and can establish better ways to scale the rewards program.

Example:

- Treasury Contract.1 Per Partner + Per Chain (where Ocean is a Tier 0 partner) that maintains all funds to be dispensed
- Balance/Mappings Contract. One per chain, that maintains all funds that were distributed to each LP
- Executor Contract. Wires Treasury & Balance Contract in order to deliver all functionality

I hold strong conviction WRT removing ownership/overhead from our team in order to execute ops in the DF program

We can delegate all responsibilities/functions/SLA to the Treasury + Executor Contract, and create the I/O for partners to integrate more freely and with confidence

(My descriiption is a bit handwavy but I think it communicates the overall concept)

----

Berkay:

The flow will be the same. The contract which keeps the mappings etc will call the functions from the other replaceable contract

Would make it much easier to migrate

trizin:
Fixes #137 

Changes proposed in this PR:

- Refactored the main contract and implemented a strategy contract.

### DIFF
--- a/contracts/DFRewards.sol
+++ b/contracts/DFRewards.sol
@@ -1,5 +1,5 @@
-pragma solidity 0.8.12;
 // SPDX-License-Identifier: (Apache-2.0 AND CC-BY-4.0)
+pragma solidity 0.8.12;
 
 import "OpenZeppelin/openzeppelin-contracts@4.2.0/contracts/token/ERC20/IERC20.sol";
 import "OpenZeppelin/openzeppelin-contracts@4.2.0/contracts/token/ERC20/utils/SafeERC20.sol";
--- a/contracts/DFStrategyV1.sol
+++ b/contracts/DFStrategyV1.sol
@@ -1,5 +1,5 @@
-pragma solidity 0.8.12;
 // SPDX-License-Identifier: (Apache-2.0 AND CC-BY-4.0)
+pragma solidity 0.8.12;
 
 import "OpenZeppelin/openzeppelin-contracts@4.2.0/contracts/token/ERC20/IERC20.sol";
 import "OpenZeppelin/openzeppelin-contracts@4.2.0/contracts/token/ERC20/utils/SafeERC20.sol";
--- a/contracts/interfaces/IPool.sol
+++ b/contracts/interfaces/IPool.sol
@@ -1,3 +1,6 @@
+// SPDX-License-Identifier: (Apache-2.0 AND CC-BY-4.0)
+pragma solidity 0.8.12;
+
 interface IPool {
     function getBaseTokenAddress() external view returns (address);
 


contracts/StrategyV1.sol
```diff
@@ -0,0 +1,98 @@
+// SPDX-License-Identifier: (Apache-2.0 AND CC-BY-4.0)
```
trentmc, line 1: Rename from "StrategyVX" to "DFStrategyVX" for consistency with "DFRewards"

contracts/DFRewards.sol
```diff
@@ -114,6 +109,15 @@ contract DFRewards is Ownable, ReentrancyGuard {
         IERC20(_token).transfer(msg.sender, amount);
     }
 
+    function addStrategy(address _strategy) external onlyOwner {
+        // maybe add a timeout here
```
trentmc, line 113: Why "maybe add a timeout"? 

I suggest: either do it, or don't, but don't comment with "maybe do x"

contracts/DFRewards.sol
```diff
@@ -1,20 +1,19 @@
...
+contract DFRewards is Ownable, ReentrancyGuard, IDFRewards {
     using SafeERC20 for IERC20;
 
-    event Allocated(address[] tos, uint256[] values, address tokenAddress);
-    event Claimed(address to, uint256 value, address tokenAddress);
-
     // token address => user address => balance
     mapping(address => mapping(address => uint256)) balances;
     mapping(address => uint256) allocated;
+    mapping(address => bool) strategies;
```
trentmc, line 16: Perhaps 'live_strategies' is a better label than 'strategies'?

And, should it be simply a _list_ of live strategies, rather than a boolean mapping?

contracts/StrategyV1.sol
```diff
@@ -0,0 +1,98 @@
...
+pragma solidity 0.8.12;
+
+import "OpenZeppelin/openzeppelin-contracts@4.2.0/contracts/token/ERC20/IERC20.sol";
+import "OpenZeppelin/openzeppelin-contracts@4.2.0/contracts/token/ERC20/utils/SafeERC20.sol";
+import "OpenZeppelin/openzeppelin-contracts@4.2.0/contracts/security/ReentrancyGuard.sol";
+import "./interfaces/IDFRewards.sol";
+import "./interfaces/IPool.sol";
+
+contract DFStrategy is ReentrancyGuard {
```
trentmc, line 10: Why "DFStrategy" instead of "DFStrategy1" ?

trentmc (changes requested):
In general, this looks great! High level approach is the most important thing, and you got it, and I like it.

My other comments are given inline, pretty small things.

contracts/DFRewards.sol
trentmc: Re live_strategies - I see you changed it, great!

contracts/StrategyV1.sol
trentmc: I see it's changed, great!

contracts/DFRewards.sol
trentmc: I'm in favor of the first one too.

contracts/StrategyV1.sol
trentmc: I see this is fixed now, great!

PR #165, opened 2022-06-22
Convert shares to basetoken

### LINKED ISSUE #160: df-py must convert pool shares to base token
The value of one pool share varies.

https://github.com/oceanprotocol/df-py/blob/main/util/query.py#L159

Here, df-py must convert the shares to base token.

trizin:
Fixes #160

Changes proposed in this PR:

- Convert shares to basetoken


trentmc:
> ```
> util/test/test_query.py ..
> ```
> 
> Query tests pass

If query tests pass, it's a sign that they weren't precise enough

PR #197, opened 2022-07-04
Filter out rewards smaller than 0.001

### LINKED ISSUE #196: Filter out rewards smaller than 0.001 OCEAN
When calculating the rewards, filter out the rewards smaller than 0.001 OCEAN

trizin:
Fixes #196 

Changes proposed in this PR:

- [Filter out rewards smaller than 0.001](https://github.com/oceanprotocol/df-py/commit/a38787ec699188e60f8afbaf8dbd6755aed0dcf6)
- [Remove the previous fix](https://github.com/oceanprotocol/df-py/commit/23895cdeb143407a6362f792c4fa2705c59c9186)

### DIFF
--- a/util/calcrewards.py
+++ b/util/calcrewards.py
@@ -192,7 +192,7 @@ def _calcRewardsUsd(
     # remove small amounts
     for chainID in chainIDs:
         for LP_addr, reward in rewardsperlp[chainID].items():
-            if rewardsperlp[chainID][LP_addr] < 0.0001:
+            if rewardsperlp[chainID][LP_addr] < 0.00001:
                 del rewardsperlp[chainID][LP_addr]
 
     # scale rewards


util/calcrewards.py
```diff
@@ -172,6 +172,8 @@ def _calcRewardsUsd(
                 if Sij == 0 or Cj == 0:
                     continue
                 RF_ij = Sij * Cj  # main formula!
+                if RF_ij < 0.001:
+                    continue
```
trentmc, line 176: Filtering won't work well here, since the units are a bit hard to reason about, being it's (stake in USD) * (volume in USD). 

I recommend instead to delete your change, and instead change "normalize and scale" part to something like the following.

1. Normalize
2. Then filter on normalized values, where units are known (a percentage)
3. Then normalize again, and scale

That is:
```python
    # normalize rewards (don't scale)
    for chainID in chainIDs:
        for LP_addr, reward in rewardsperlp[chainID].items():
            rewardsperlp[chainID][LP_addr] = reward / tot_rewards

    #filter out any normalized reward that's less than 0.001% (=0.00001)
    for chainID in chainIDs:
        for LP_addr, reward in rewardsperlp[chainID].items():
            if rewardsperlp[chainID][LP_addr] < 0.00001:
                del rewardsperlp[chainID][LP_addr]

    # normalize again, *and* scale rewards
    for chainID in chainIDs:
        for LP_addr, reward in rewardsperlp[chainID].items():
            rewardsperlp[chainID][LP_addr] = reward / tot_rewards * TOKEN_avail
```

util/dispense.py
```diff
@@ -46,7 +46,6 @@ def dispense(
 
     to_addrs = list(rewards.keys())
     values = [toBase18(rewards[to_addr]) for to_addr in to_addrs]
-    values = [max(value, 1) for value in values]  # avoid zero
```
trentmc, line 49: Good idea to remove this, it forces us to do things cleaner upstream

util/calcrewards.py
trentmc: Whups, yes, do that.

trentmc (approved):

PR #217, opened 2022-07-11
Remove `st_block` from the seed

### LINKED ISSUE #185: Reward calculations inconsistent run-to-run, despite new random seed functionality
**Describe the bug**
Towards doing payment for DF2, Berkay ran one run to calc LP rewards, and Trent ran two runs. Here's the data:
https://drive.google.com/drive/folders/1Aj_E5OQzCLtHMthGpTogrbwQIWW1Y_ZA

Results in terms of OCEAN payout to LP for all three were similar, but not quite the same. It's like different random seeds were used, therefore different block samples were chosen. (Even though the new functionality is supposed to have a random seed)

**To Reproduce**

To create new data: do the steps at https://github.com/oceanprotocol/df-py/blob/main/README-dist-ops.md, for Jun 23-30. (Timeframe of DF2). Use the DF2 random seed.

**Hypotheses**

- Hyp 1: random seed not working end-to-end
- Hyp 2: when we calculate using "now", even though unix says "june 30", if a few minutes or blocks pass, then perhaps it's using up all the blocks today. When it really should have ended at the block of last night.

trizin:
Fixes #185 - probably

Changes proposed in this PR:

- Remove `st_block` from the seed.

### DIFF
--- a/dftool
+++ b/dftool
@@ -140,8 +140,7 @@ SECRET_SEED -- secret integer used to seed the rng
 
     # main work
     recordDeployedContracts(ADDRESS_FILE)
-    seed = st_block + SECRET_SEED
-    rng = blockrange.BlockRange(st_block, fin_block, NSAMP, seed)
+    rng = blockrange.BlockRange(st_block, fin_block, NSAMP, SECRET_SEED)
     rng.filterByMaxBlock(len(chain))
     (Pi, Si, Vi, Ai, SYMi) = query.query_all(rng, CHAINID)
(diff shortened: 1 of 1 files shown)


trentmc (approved):

trentmc:
> Fixes #185 - probably

Before you merge: can you check to see if it does this?

trentmc:
> I can't reproduce the issue on my machine

Alright. I guess just merge. And move on:)

PR #224, opened 2022-07-18
Smart remove allocation

trizin:
Changes proposed in this PR:

- Removing 100% allocation swaps the last element with the element id. Thus there won't be records with zero allocation.
- Added `limit` and `skip` params to `totalAllocation` function.
- Added new mapping `idToAllocation`.
- Added tests.

This might be a waste of gas, but the best solution I found in case we want to read data directly from the chain.


trentmc:
Why is this a separate PR, versus the main veOCEAN PR?

PR #231, opened 2022-07-22
dfve - Improvement/expanding test coverage

idiom-bytes:
WIP - dfve
- Implementing functionality to deploy + use FREs
- Next: Implement functionality to randomly lock veOCEAN, consume data, etc...

### DIFF
--- a/util/oceantestutil_ve.py
+++ b/util/oceantestutil_ve.py
@@ -1,7 +1,9 @@
 import random
-import brownie
 
+import brownie
+from brownie.convert.main import to_uint
 from enforce_typing import enforce_types
+
 from util import constants, oceanutil
 from util.base18 import toBase18, fromBase18
 
@@ -33,6 +35,7 @@ def deployDataNFTWithFRE(from_account, token):
     exchangeId = oceanutil.createFREFromDatatoken(
         DT,
         token,
+        10.0,
         from_account
     )
 
--- a/util/oceantestutil_ve.py
+++ b/util/oceantestutil_ve.py
@@ -1,0 +1,39 @@
+import random
+import brownie
+
+from enforce_typing import enforce_types
+from util import constants, oceanutil
+from util.base18 import toBase18, fromBase18
+
+network = brownie.network
+
+@enforce_types
+def randomDeployFREsThenConsume(num_FRE: int, base_token):
+    accounts = network.accounts
+
+    # create random num_FRE.
+    tups = []  # (pub_account_i, data_NFT, DT, FRE)
+    for fre_i in range(num_FRE):
+        if fre_i < len(accounts):
+            account_i = fre_i
+        else:
+            account_i = random.randint(0, len(accounts))
+        (data_NFT, DT, FRE) = deployDataNFTWithFRE(accounts[account_i], base_token)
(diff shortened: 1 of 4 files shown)


util/oceantestutil_ve.py
```diff
@@ -0,0 +1,42 @@
+import random
```
trentmc, line 1: Suggest: just put this into oceantestutil.py, not a new file

util/oceantestutil_ve.py
```diff
@@ -0,0 +1,42 @@
...
+
+import brownie
+from brownie.convert.main import to_uint
```
trentmc, line 4: `to_uint` isn't being used, not sure why imported. In this file and other files.

util/oceanutil.py
```diff
@@ -84,6 +85,18 @@ def ERC721Factory():
     return _contracts("ERC721Factory")
 
 
+def veOCEAN():
+    return _contracts("veOCEAN")
+
+
+def veAllocate():
+    return _contracts("veOCEAN")
```
trentmc, line 93: Good, this is exactly where to put this:) +1

util/test/test_query_ve.py
```diff
@@ -0,0 +1,89 @@
+import time
```
trentmc, line 1: In this file, move test_populateSubgraph() into test_query.py, as a _test_populateSubgraph(). Why: avoids an extra 30 s of loops, and the code for loops you needed to add; instead it reuses the infrastructure already set up in test_query.py.

util/test/test_query_ve.py
```diff
@@ -0,0 +1,89 @@
...
+    C = CONTRACTS[CHAINID]
+    C["Ocean"] = ocean
+    C["ERC721Template"] = templateERC721
+    C["ERC20Template"] = templateERC20
+    C["PoolTemplate"] = poolTemplate
+    C["Router"] = factoryRouter
+    C["ERC721Factory"] = factoryERC721
+    C["FixedPrice"] = fixedRateExchange
+    C["veOCEAN"] = ve_ocean
+    C["veAllocate"] = ve_allocate
```
trentmc, line 85: You shouldn't need any of this, because [oceanutil.py already does this](https://github.com/oceanprotocol/df-py/blob/5fd40fdbc2a0336a63c59a5dd00250be76c98806/util/oceanutil.py#L43).

And like I recommend above, you can delete this whole file, and move the test function itself into test_query.py. Therefore this file's setup_function() gets deprecated.

trentmc (changes requested):

idiom-bytes:
Went to a place w/ no internet. Was able to work w/ everything local. Great.

WIP, but pretty close:
- createDataNFTWithFREs
- randomlyCreatelocks
- randomlyAllocate
- randomlyConsume

Working through a few, final kinks w/ passing arrays to contracts while using specific data types.

trentmc:
> I'm stuck at this point. I can't figure out why the `exchangeId` is invalid.

Were you able to work through the FRE readme first?  https://github.com/oceanprotocol/ocean.py/blob/main/READMEs/marketplace-flow.md

PR #254, opened 2022-08-17
Update veAllocate to max 100 points across all dataNFT

### LINKED ISSUE #252: Update veAllocate to max 100 points per dataNFT
Everything else can be handled on the FE side

trizin:
Fixes #252 

Changes proposed in this PR:

- Update `veAllocate` to max 100**0** points across all dataNFT

I set the max to 1000, so we can have more precision.


idiom-bytes:
I thought this was going to be 10,000 so we could have 2 decimals?

Besides that, LGTM!

PR #292, opened 2022-09-15
Filter non-market assets

### LINKED ISSUE #291: Filter non-market assets - When serving assets, If there is no DDO, do not serve it
Problem:
We're getting assets from Accentrik in our market, although there is no way to properly find their DDO or connect to their asset.

Required Steps:
1. Please query all assets available
2. Filter out assets that do not have a DDO
3. Only serve assets to the user for them to allocate, that have a DDO available
4. All queries should be limited by this way, so we only serve available DDOs

This should filter out the basic assets that should not be available to users.

trizin:
Fixes #291 

Changes proposed in this PR:

- Added `getAquariusAssetNames` function. Given an array of dids returns a mapping of did -> asset name.
- Added a test for `getAquariusAssetNames` function.
- Added `_filterOutNonMarketAssets` function. Filters out non-market assets from the `nftvols` dict.
- Added a test for `_filterOutNonMarketAssets` function.
- Filtered out non-market assets in the `query_all` function. (only filters on production)

Further work:
- [ ] We should find a way to test the filtering of non-market assets in the `query_all` function by perhaps leveraging Aquarius in the barge. We will also need to "register" all generated assets to Aquarius in other tests that use `query_all`. Or perhaps we could mock `getAquariusAssetNames` function.
- [X] I need to learn if there is a limit for the Aquarius endpoint. If that's the case, I'll adjust the batch size. Edit: The batch size is around 10k.

### DIFF
--- a/util/query.py
+++ b/util/query.py
@@ -369,7 +369,7 @@ def _filterNftvols(nftvols: dict, chainID: int) -> dict:
         # can't filter on dev chain:
         return nftvols
 
-    filtered_nfts: Dict[str, Dict[str, float]] = {}
+    filtered_nftvols: Dict[str, Dict[str, float]] = {}
     nft_dids = []
 
     for basetoken_addr in nftvols:
@@ -382,13 +382,13 @@ def _filterNftvols(nftvols: dict, chainID: int) -> dict:
         for nft_addr in nftvols[basetoken_addr]:
             did = oceanutil.calcDID(nft_addr, chainID)
             if did in filtered_dids:
-                if basetoken_addr not in filtered_nfts:
-                    filtered_nfts[basetoken_addr] = {}
-                filtered_nfts[basetoken_addr][nft_addr] = nftvols[basetoken_addr][
+                if basetoken_addr not in filtered_nftvols:
+                    filtered_nftvols[basetoken_addr] = {}
+                filtered_nftvols[basetoken_addr][nft_addr] = nftvols[basetoken_addr][
                     nft_addr
                 ]
 
-    return filtered_nfts
+    return filtered_nftvols
 
 
 @enforce_types


util/query.py
```diff
@@ -327,7 +333,7 @@ def getNFTVolumes(
 def _filterOutPurgatory(nftvols: dict, chainID: int) -> dict:
```
trentmc, line 333: It appears that _filterOutPurgatory() was broken before your changes, since at each iteration of the loop it's updating the whole entry for basetoken, which could be a lot of NFTs. 

So this needs to get fixed, where it only updates the entry inside the basetoken dict. You can do that as part of this PR. I see that you're doing something like that in the new function _filterOutMarketAssets(). Great, use that pattern here.

util/query.py
```diff
@@ -336,12 +342,47 @@ def _filterOutPurgatory(nftvols: dict, chainID: int) -> dict:
...
         for nft_addr in nftvols[basetoken_addr]:
             if oceanutil.calcDID(nft_addr, chainID) not in bad_dids:
-                filtered_pools[basetoken_addr] = nftvols[basetoken_addr]
-    return filtered_pools
+                filtered_nfts[basetoken_addr] = nftvols[basetoken_addr]
+    return filtered_nfts
+
+
+@enforce_types
+def _filterOutNonMarketAssets(nftvols: dict, chainID: int) -> dict:
```
trentmc, line 354: I'd re-label from "_filterOutNonMarketAssets" to "_filterToAquariusAssets". Why:
- "non-market" is vague. Aquarius is precise
- replace the double-negative ("out" and "non-market") with a positive ("just aquarius")

util/query.py
```diff
@@ -336,12 +342,47 @@ def _filterOutPurgatory(nftvols: dict, chainID: int) -> dict:
...
+      Return nfts that belong to the Ocean marketplace
+
+    @arguments
+      nftvols: dict of [basetoken_addr][nft_addr]:vol_amt
+
+    @return
+      filtered_nftvols: list of [basetoken_addr][nft_addr]:vol_amt
+    """
+    filtered_nfts: Dict[str, Dict[str, float]] = {}
+    didList = []
```
trentmc, line 366: Problems: "didList" label is too vague, and camelCase. 

Fix by: rename to nft_dids. This parallels "nftvols".

Finally, it should be a set not a list, to make clear that there might be duplicates.

util/query.py
```diff
@@ -336,12 +342,47 @@ def _filterOutPurgatory(nftvols: dict, chainID: int) -> dict:
...
+
+    for basetoken_addr in nftvols:
+        for nft_addr in nftvols[basetoken_addr]:
+            didList.append(oceanutil.calcDID(nft_addr, chainID))
+
+    aquariusAssetNames = getAquariusAssetNames(didList)
```
trentmc, line 372: `aquariusAssetNames` -> `aquarius_asset_names` here, and below. Snake case not camel case.

`didList` -> `list(nft_dids)` to be in line with above.

trentmc (changes requested):

util/constants.py
```diff
@@ -3,6 +3,7 @@
 ZERO_ADDRESS = "0x0000000000000000000000000000000000000000"
 
 MARKET_ASSET_BASE_URL = "https://market.oceanprotocol.com/asset/"
```
trentmc, line 5: It appears that `MARKET_ASSET_BASE_URL` is unused? If yes, you can delete the line.

(It's not really this PR, however it's ok if we remove it here because it's kinda related)

trentmc (changes requested):
Great fixes to my comments!

I made one small suggestion. After that, I'll be ready to approve

util/query.py
```diff
@@ -324,24 +325,94 @@ def getNFTVolumes(
...
     @return
       filtered_nftvols: list of [basetoken_addr][nft_addr]:vol_amt
     """
-    bad_dids = _didsInPurgatory()
-    filtered_pools = {}
+    if chainID == networkutil.DEV_CHAINID:
+        # can't filter on dev chain:
+        return nftvols
+
+    filtered_nfts: Dict[str, Dict[str, float]] = {}
```
trentmc, line 372: This overall update looks great! Super clean.

One small thing: rename `filtered_nfts` to `filtered_nftvols`. 

Then, good to go!

trentmc (approved):

PR #297, opened 2022-09-16
Include all assets in the `nftinfo` csv

### LINKED ISSUE #296: Nftinfo csv should include all assets
Nftinfo csv should include all assets

trizin:
Fixes #296 

Changes proposed in this PR:

- Separated nft volumes and nft info.
- Added a new function to get the list of all nfts and a test.

### DIFF
--- a/util/query.py
+++ b/util/query.py
@@ -235,6 +235,24 @@ def getAllocations(
 def getNFTInfos(chainID) -> List[DataNFT]:
     """
     @description
+      Fetch, filter and return all NFTs on the chain
+
+    @return
+      nftInfo -- list of DataNFT objects
+    """
+
+    NFTinfo = _getNFTInfos(chainID)
+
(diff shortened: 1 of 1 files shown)


idiom-bytes (approved):
I reviewed it from a high level, and am familiar with the other comments by trent. Will dig into it myself as I try to run the data pipeline E2E. 

Ran locally, completed successfully.

trentmc (changes requested):
Small changes requested.

I'm a bit uneasy with one more query - since do error prone - but I don't see a better way.

Query.py's code flow for filtering takes a bit of work to decipher. But it's not horrible. However if you see ways to help devs understand this flow better - in a maintainable way - go for it. I'd suggest adding a fewer lines of documentation in the code, but I fear that will be hard to maintain.

One thing that may help: is there a way to split query.py into two (or three) files? It might solve my concerns above.

PR #304, opened 2022-09-20
Support free assets (in DF rewards)

### LINKED ISSUE #298: Support free assets, by using gas cost for price measure. So DF incentivizes for Binance price feed etc
**Is your feature request related to a problem? Please describe.**

It's useful if DF can incentivize for free assets. Why: it greatly increases the Ocean use cases that DF incentivizes for, since there's 100x+ more free data than priced data. Specifically, it will ensure there's DF incentives for Binance API price feeds, etc.

But, there are challenges:
- DF rewards based on price * (# consumes). But free assets have price = 0, by definition
- Wash consuming. Our workaround for priced assets is to charge a consume fee % against price, and to understand that wash trading becomes unprofitable once enough consumes are done; it's a threshold that can get surpassed quickly. But for free assets, since price = 0, then consume fee = 0

**Candidate Solutions**

**Idea 0. Have a bonding curve for free datasets.** 
- Con: quite complex. We'd had this idea when we still had AMMs for datatokens, and DF "staking on a dataset" was for the AMM, therefore the BC was the staking equivalent for free datasets. But now, veOCEAN takes care of staking. So the problem is simpler to solve, and we probably don't need to resort to BCs. So what other ideas are there? ... 

**Idea 1. For whole Ocean stack, have a price of e.g. $0.01 for all "free" assets.** 
- Con: this hurts several use cases where there should be many, many shares. 
- Con: this affects the whole Ocean stack; significant changes may be needed (eg GUI for "free" price)

**Idea 2. In DF, when we measure price, then measured_price = max($0.01, price seen).** That way free assets and super-cheap assets have a lower-bound price.
- Pro: only affects DF stack. 
- Con: very bad wash consume. That is: in free assets, if price > gas fees, then DF users could wash consume until they get all DF rewards for themselves.

**Idea 3. In DF, when we measure price, then measured price = price seen + gas cost to consume.** So even free assets have a price.
- Pro: only affects DF stack
- Pro: no wash consume on free assets. It's unprofitable.

**Describe the solution you'd like**

Use idea 3: In DF, when we measure price, then measured price = price seen + gas cost to consume. So even free assets have a price.

trizin:
Fixes #298

Changes proposed in this PR:

- Support free assets

### DIFF
--- a/util/networkutil.py
+++ b/util/networkutil.py
@@ -10,30 +10,23 @@ _BARGE_ADDRESS_FILE = "~/.ocean/ocean-contracts/artifacts/address.json"
 # Development chainid is from brownie, rest are from chainlist.org
 # Chain values to fit Ocean subgraph urls as given in
 # https://v3.docs.oceanprotocol.com/concepts/networks/
-_CHAINID_TO_NETWORK = {
-    8996: "development",  # ganache
-    1: "mainnet",
-    3: "ropsten",
-    4: "rinkeby",
-    56: "bsc",
-    137: "polygon",
-    246: "energyweb",
-    1287: "moonbase",
-    1285: "moonriver",
-    80001: "mumbai",
-}
-_CHAINID_TO_NATIVE_TOKEN = {
-    8996: "OCEAN",
-    1: "ETH",
-    3: "ETH",
-    4: "ETH",
-    56: "MATIC",
-    137: "BNB",
-    246: "EWT",
-    1287: "MOVR",
-    1285: "MOVR",
-    80001: "MATIC",
-}
+
+_RAW_CHAIN_DATA = [
+    (8996, "development", "OCEAN"),
+    (1, "mainnet", "ETH"),
+    (3, "ropsten", "ETH"),
+    (4, "rinkeby", "ETH"),
+    (56, "bsc", "MATIC"),
+    (137, "polygon", "BNB"),
+    (246, "energyweb", "EWT"),
+    (1287, "moonbase", "MOVR"),
+    (1285, "moonriver", "MOVR"),
+    (80001, "mumbai", "MATIC"),
+]
+
+_CHAINID_TO_NETWORK = {x[0]: x[1] for x in _RAW_CHAIN_DATA}
+_CHAINID_TO_NATIVE_TOKEN = {x[0]: x[2] for x in _RAW_CHAIN_DATA}
+
 
 _NETWORK_TO_CHAINID = {
     network: chainID for chainID, network in _CHAINID_TO_NETWORK.items()


util/constants.py
```diff
@@ -10,3 +10,22 @@
...
+
+
+WRAPPED_TOKEN_SYMBOLS = {
+    WRAPPED_TOKEN_ADDRS[1]: "ETH",
+    WRAPPED_TOKEN_ADDRS[3]: "ETH",
+    WRAPPED_TOKEN_ADDRS[4]: "ETH",
+    WRAPPED_TOKEN_ADDRS[137]: "MATIC",
+    WRAPPED_TOKEN_ADDRS[56]: "BNB",
+    # TODO Add more
+}
```
trentmc, line 31: These might be better placed in util/networkutil.py.

Also, rather than mapping a chainid int to the value, perhaps make it chain name to the value. Less error-prone.

util/query.py
```diff
@@ -339,10 +349,21 @@ def getNFTVolumes(
...
         for order in new_orders:
             lastPriceValue = float(order["lastPriceValue"])
-            if lastPriceValue == 0:
-                continue
-            nft_addr = order["datatoken"]["nft"]["id"].lower()
             basetoken_addr = order["lastPriceToken"]
+            nft_addr = order["datatoken"]["nft"]["id"].lower()
+
+            if lastPriceValue == 0:
+                if len(order["dispensers"]) > 0:
```
trentmc, line 356: Rather than adding gas cost _only_ for free assets, I propose we add it for _all_ asset. That is, include priced assets too. Why:
- It better tracks the actual $ spent by the consumer to consume an asset. More realistic.
- It will simplify our code. E.g. here you don't need to test if "lastPriceValue == 0".

util/tousd.py
```diff
@@ -51,6 +52,10 @@ def nftvolsToUsd(
     cleancase.assertNFTvols(nftvols)
     cleancase.assertRates(rates)
 
+    # Add native token rates
+    for chainID in WRAPPED_TOKEN_ADDRS.keys():
+        symbols[chainID][WRAPPED_TOKEN_ADDRS[chainID]] = WRAPPED_TOKEN_SYMBOLS[chainID]
```
trentmc, line 57: There's too much going on in a single line. It takes too much effort to decipher. Something as simple as this should be breezy easy.

I recommend to break it into 2-3 lines. Compute the addr on one line. Compute symbol on another. And on the third, update the dict.

trentmc (changes requested):

util/networkutil.py
```diff
@@ -28,6 +28,33 @@
...
+    1: "0xeth",
+    3: "0xropsten",
+    4: "0xrinkeby",
+    56: "0xbnb",
+    137: "0xpolygon",
+    246: "0xenergyweb",
+    1285: "0xmoonriver",
+    1287: "0xmoonbase",
+    80001: "0xmumbai",
+}
```
trentmc, line 42: We can't have this, big DRY violation due to overlap with `_CHAINID_TO_NETWORK`. If you want to have this dict, auto-compute it as a function of `_CHAINID_TO_NETWORK`.

trentmc (changes requested):

trentmc (approved):

util/networkutil.py
trentmc: What if we create a list in the code, something like:

```python
_CHAIN_INFO = [
#chainid, name, native_token
8996, "development", "ETH",
1, "mainnet", "ETH",
3, "ropsten", "ETH",
...
137, "polygon", "MATIC",
...
]

Then, other caches or dicts can be computed from this.

Less error prone, because all data is in one place.

trentmc (approved):

PR #318, opened 2022-09-28
Remove approved tokens

### LINKED ISSUE #317: Remove approved tokens
Remove approved tokens and add chain native tokens to symbols csv so df-sql recognizes free assets

trizin:
Fixes #317

Changes proposed in this PR:

- Remove approved tokens.
- Add chain native tokens to symbols csv.

### DIFF
--- a/util/test/test_query.py
+++ b/util/test/test_query.py
@@ -149,7 +149,7 @@ def _test_getAllocations(rng: BlockRange):
 def _test_getSymbols():
     oceanToken = oceanutil.OCEANtoken()
     tokset = TokSet()
-    tokset.add(CHAINID, oceanToken.address, "OCEAN")
+    tokset.add(CHAINID, oceanToken.address.lower(), "OCEAN")
     symbols_at_chain = query.getSymbols(
         tokset, CHAINID
     )  # dict of [basetoken_addr] : basetoken_symbol


trentmc (changes requested):
> [Berkay] Could you take a look at this? The solution I found feels a bit wrong wanted to ask you before i update the tests

My response: correct. It's not the right place for this. Because now the meaning of "approved tokens" has been distorted to something weird.

To solve the goal of the issue, I'd get closer to the code where df-sql is doing its filtering. Eg Can it know if an asset is free or not?

trentmc:
> Makes sense but there is another thing I forgot to mention. If we don't have them in the CSV files the reward calculation won't include free assets.

I started going down the path of investigating the code that would need changing to allow this.

Then I realized: we don't _care_ what the asset is priced in anymore, at all. Why:
- Before, we did, because it was to bias to OCEAN or H2O as liquidity in AMMs. 
- But now, the token is just a unit of exchange. We don't need to bias to OCEAN or H2O as unit-of-exchange at all.
- The blog post says this too actually. (But we just never propagated to code)

Therefore: we can _deprecate_ all the approved tokens stuff! Instead, allow pricing in whatever token, it's ok.

You can do it via this issue / PR (and rename the issue + PR accordingly), or create a new issue + PR (and close the current ones)

trentmc (approved):
I only did a cursory view. Let's ship this:)

PR #338, opened 2022-10-11
Update FeeEstimate

### LINKED ISSUE #335: Update FeeEstimate to not rely on checkpoints inside FeeDistributor
### Problem:
- Right now, we're running into overhead issues because we need to checkpoint FeeDistributor daily, for the estimate calc to work
- This means that we're wasting gas and time, however when reviewing the code we believe that we can get around this by simply updating the estimateClaim() fn

### Candidate solution:
Cand A:
- We believe that we can calc the rewards without depending on checkpoints
- Update FeeEstimate to no rely on checkpoint
**- Users should be able to get an accurate reward estimate without us having to checkpoint**
Cand B:
- Automate checkpoint with gelato
Cand C:
- Add an update checkpoint in the FE and let users update the checkpoint
Cand D:
- Use a PK + Brownie to automate the call w/ a cron job

### Constraints:
- Let's start with a tight estimate around this (1-2 days) if we run into more problems and this grows in size, let's pause and re-asses

### DoD:
- [ ] FeeEstimator works and gives an accurate estimate to the FE while ignoring the FeeDistributor checkpoint

trizin:
Fixes #335

Changes proposed in this PR:

- Updated FeeEstimate so we don't have to call the checkpoint function every 24 hours.

### DIFF
--- a/util/test/veOcean/test_fee_estimate.py
+++ b/util/test/veOcean/test_fee_estimate.py
@@ -1,0 +1,130 @@
+import brownie
+from enforce_typing import enforce_types
+import pytest
+
+from util import networkutil, oceanutil
+from util.constants import BROWNIE_PROJECT as B
+from util.base18 import toBase18, fromBase18
+from datetime import datetime
+
+accounts = None
+alice = None
+bob = None
+veOCEAN = None
+OCEAN = None
+DAY = 86400
+WEEK = 7 * 86400
+MAXTIME = 4 * 365 * 86400  # 4 years
+chain = brownie.network.chain
+TA = toBase18(10000.0)
+DAY = 86400
+
+
+@enforce_types
+def test_rewards():
+    t0 = chain.time()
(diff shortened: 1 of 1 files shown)


alexcos20:
I don't think it's so easy..
Can you create a test flow like:
 - deploy veOcean and feeDistributor
 - Alice locks OCEAN, Bob locks Ocean
 - make sure estimate returns 0, no reverts
 - advance to the next week, without any checkpoints
 - top-up feeDistributor with some rewards, call checkpoint_token & checkpoint_total_supply
 - advance 25 hours, without any checkpoints
 - make sure estimate return 0, no reverts
 - advance to next week, without any checkpoints
 - make sure estimates returns > 0
 - Alice claims , make sure that claimed amount == estimated_reward from previous step
 - advance 25 hours, without any checkpoints
 - make sure that Bob estimate_rewards >0 , no reverts
 - make sure that Alice estimate_rewards = 0 , no reverts

Thanks

PR #340, opened 2022-10-19
Fix allocation calc

trizin:
Changes proposed in this PR:

- Fix allocation calc.
- [Allow for rounding errors](https://github.com/oceanprotocol/df-py/pull/340/commits/04de8ebfba592c47a4a3cc8b1376519059d429af)
- [Normalize values](https://github.com/oceanprotocol/df-py/pull/340/commits/d20236a5f2b0f4ac05560d0cc9bfebec5bc4b1a0)

### DIFF
--- a/util/query.py
+++ b/util/query.py
@@ -281,7 +281,7 @@ def getAllocations(
         for nft_addr in allocations[chain_id]:
             for LP_addr in allocations[chain_id][nft_addr]:
                 if lp_total[LP_addr] == 0.0:
-                    print(f"WARNING: lp_total[LP_addr] == 0.0")
+                    print(f"WARNING: {lp_total[LP_addr]} == 0.0")
                     continue
                 allocations[chain_id][nft_addr][LP_addr] /= lp_total[LP_addr]
 


util/query.py
```diff
@@ -264,6 +263,28 @@ def getAllocations(
                 n = num_allocations[chain_id][nft_addr][LP_addr]
                 allocations[chain_id][nft_addr][LP_addr] /= n
 
+    # get total allocations per each LP
+    lp_total = {}
```
trentmc, line 267: Overall, this PR looks like it's doing what it's supposed to do. So my feedback below is to improve the maintainability of the code.
---
There's too much going on in getAllocations(). 

It really should be two methods:
1. get raw allocations
2. normalize raw allocations

- Step (2) shouldn't really be called in query.py, it should be called from some higher-level routine. 
- Step (2) is more naturally done as numpy arrays, like in calcrewards.py. It will be less error-prone.

idiom-bytes: @trentmc i see what you are saying but not every math operation on an array needs to be done in numpy because it's python

wrt to df_rewards, we are setting up a workflow that connects contracts, backends, apis, etc...
it's easy to read and engineer-friendly

the loops/entry could perhaps be done in a more pythonic way, but I think it fits the already established structure of what is acceptable in df-py.

It's 20+ lines to solve the issue and it's well contained to that function.
**I agree on perhaps moving it to a higher-level routine.**

In this case, I propose writing a ticket, and we can improve it as we go.
Let's get the end-to-end done.

idiom-bytes: I reviewed query, and I believe this also fits in there.
It has been growing in size, and it should likely get trimmed down at some point.

trentmc: FYI Berkay and I have been DM'ing. He merged the loopy version of the commit. Separately, he will try a numpy version. I'll summarize the conversation here.

"Being pythonic" != "forcing the use of loops". 

numpy is not just another library, it's one of the most widely used Python libraries. M's or tens or M's of people use it. It's at the heart of the numerical python community. 

> the loops/entry could perhaps be done in a more pythonic way, but I think it fits the already established structure of what is acceptable in df-py.

df-py uses loops where appropriate, and numpy where appropriate. Eg calcrequests.py uses numpy because it's appropriate there. What would have been 3- or 4-deep nested loops is now single lines via arrays. Same thing for here for normalize raw allocations.

idiom-bytes: i understand they are not the same, i never meant to convey they were
due to the requirement of applying math over arrays, it totally makes sense

but, if it can be solved by a lambda/map function (i.e. a pythonic approach) so an engineer can solve it in 30 minutes rather than spending 1 day hacking it away in numpy, I would say: "go for it" (get the e2e done)

trentmc: If you can think in arrays, it's 30 min in numpy.

I repeat: numpy is also pythonic.

The end result is going to be easier to read and maintain, because the problem at hand is naturally arrays.

PR #342, opened 2022-10-20
Fix sampling

trizin:
Previous method only samples if the user has existing allocation or ve balance. This PR should fix it.


idiom-bytes (approved):
Looks good, thanks for catching this berkay. let's keep improving.

trentmc (approved):

PR #352, opened 2022-10-27
Block finder #336

### LINKED ISSUE #336: Reward Calc should report consistent results across validators

trizin:
Fixes #336

Changes proposed in this PR:

- Added functions `ethTimestamptoBlock` and `ethCalcBlockNumber` which estimates the block number from timestamp.
- Added a function `ethFindFirstThuBlock` which finds first block in Thursday closest to the given block number.

### DIFF
--- a/dftool
+++ b/dftool
@@ -1,6 +1,6 @@
 #!/usr/bin/env python
 # pylint: disable=too-many-lines
-# we should divide this file into submodules in the future
+import datetime
 import os
 import sys
 import functools
@@ -11,14 +11,14 @@ from web3.middleware import geth_poa_middleware
 from util import (
     allocations,
     blockrange,
-    networkutil,
+    calcrewards,
     csvs,
     dispense,
     getrate,
+    networkutil,
     query,
 )
 from util.base18 import fromBase18
-from util.calcrewards import calcRewards
 from util.constants import BROWNIE_PROJECT as B
 from util.networkutil import DEV_CHAINID
 from util.oceanutil import (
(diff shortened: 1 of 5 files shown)


trentmc:
What github issue is this for, please? (It should always be in the PR description. And I find it easier if it's in the title too.)

util/test/test_blocktime.py
```diff
@@ -147,14 +153,63 @@ def test_getstfinBlocks():
     assert fin == 0
 
 
+@enforce_types
+def test_ethTimestamptoBlock():
+    networkutil.connect(1)
```
trentmc, line 158: Every unit test in this module automatically calls `setup_function()` before doing the body of the test. Inside `setup_function()`, it calls `networkutil.connect()`. 

Every unit test also auto-calls `teardown_function()` after doing the body of the test. Inside `teardown_function()`, it calls `networkutil.disconnect()`. 

Therefore:
- why does this test (`test_ethTimestamptoBlock`) have `networkutil.connect(1)` in the first line? (If it does need it, it should pass in `networkutil.DEV_CHAINID` not `1`.
- and, why does it again have `networkutil.connect(networkutil.DEV_CHAINID)` in the last line?

util/test/test_blocktime.py
```diff
@@ -147,14 +153,63 @@ def test_getstfinBlocks():
...
+
+    guess = ethTimestamptoBlock(_chain, ts)
+
+    assert guess == approx(block, 10)
+
+    networkutil.connect(networkutil.DEV_CHAINID)
+
+
+@enforce_types
+def test_ethFindFirstThuBlock():
```
trentmc, line 172: For this test, I have the same questions about network setup and teardown as in the previous test.

util/blocktime.py
```diff
@@ -133,6 +133,68 @@ def timeSinceTimestamp(self, block_i):
...
+        block_number -- int
+    @description
+        Finds the first Thursday block closest to the given block number
+    """
+
+    def get_tsdtday(block_i):
+        block_ts = chain[block_i].timestamp
+        block_dt = datetime.fromtimestamp(block_ts)
+        block_day = block_dt.weekday()
+        return block_ts, block_dt, block_day
```
trentmc, line 175: This function returns three items, though all of its callers only use the third item.

Also, the function name is cryptic (a concatenation of all return items?). Probably to fix: only return the third item, then rename accordingly (and less cryptic).

util/test/test_blocktime.py
```diff
@@ -147,14 +153,63 @@ def test_getstfinBlocks():
...
 @enforce_types
 def teardown_function():
     networkutil.disconnect()
+
+    global PREV
+
+    if PREV.WEB3_INFURA_PROJECT_ID is None:
+        del os.environ["WEB3_INFURA_PROJECT_ID"]
+    else:
+        os.environ["WEB3_INFURA_PROJECT_ID"] = PREV.WEB3_INFURA_PROJECT_ID
```
trentmc, line 215: Macro-level comment: if the changes here render other methods obsolete (because we don't use them anymore), then please delete them.

trentmc (changes requested):

trentmc (approved):

PR #396, opened 2022-12-13
Add `gen_hist_data` script

trizin:
Changes proposed in this PR:

- `gen_hist_data.sh`
Generates all csv files for the specified round number.
Usage:
```
$ gen_hist_data.sh {round_number} {output_folder}
```

# DoD:
- [x] script that generates all the csvs for historical_data
- [x] write a test that validates this command is working (dftool test)


trentmc (changes requested):
I'd like to see an automated test for each of these scripts.

And, I wonder if the body of this should be in a python script rather than a bash script. It may be less error prone. And probably easier to test. (But I'm on the fence.)

idiom-bytes:
Moved additional shell scripts and tests that will be created, in this other ticket.
https://github.com/oceanprotocol/df-py/issues/491

Am going to disable the shell test since it's passing.
The time to execute the shell script (poor estimate) is ~18m.

trentmc (changes requested):
I thought more about this script.

Having it as one big .sh makes it much harder to validate.

I recommend: push the functionality into python, where things are far more testable. 

Specifically: 
- use 'dftool' as last-mile CLI interface, rather than a new .sh. Eg `dftool genhistdata`
- inside `dftool::do_genhistdata()` have a pass-through to a new module `util/histdata.py`
- `histdata.py` makes calls to getrate(), query() etc. But the calls are directly to the _Python_ functions, _not_ to the CLI version. (Remember, all of df-py is like this: for any CLI command you see, it's a thin wrapper for some Python function.)
- then test it in Python unit tests
  - @idiom-bytes had commented elsewhere that doing unit tests was too slow before, because of a query to Polygon, and recommended turning off the test. 
  - Better: _keep_ the test but simply set it up such that you're only using Ganache, not Polygon
  - Where to put the test: _not_ `test_histdata.py`, because that would make populating Ganache super-slow. Rather, add to `test_query.py::test_all()` which reuses the "populate Ganache" action for many modules

idiom-bytes:
```
use 'dftool' as last-mile CLI interface, rather than a new .sh. Eg dftool genhistdata
```
I feel like this is the case already, genhistdata is a lot of configs, params, and functions stitched together.
Same will all.sh and nftinfo.sh

```
Where to put the test: not test_histdata.py, because that would make populating Ganache super-slow. Rather, add to test_query.py::test_all() which reuses the "populate Ganache" action for many modules
```
I guess my question is:
- Do we just want to test all the functions in 1 complex example like: test_query.py::test_all() and then have many scripts that just use df-py as a base library?
Or
- Continue to expand every functionality as part of the library?

df-py as-is, provides a lot of great functionality by stitching a bunch of macro commands together. I believe it would be great for us to continue to explore `scripts/` and `.yml` type jobs rather than go very granular with the python library.

Maybe that's a better question to ask:
1. What should be library `df-py` code vs. `script` code
2. What can our exploration process > ossification be like. Does it have to be tight? Or can it be loose?

Perhaps is better to continue explore what's possible in `script` code, and in 1-2Qs, ossify it into the library?

Personally, I'd love to add all.sh, nftinfo.sh, before moving this into histdata.py (I have pasted them below)

**Key Point:** I am trying to close this ticket, so we can close the sprint.
Changing it into python, and completing the rest of this work, does not have to be attached to this task.

```
all.sh
=========================
#rm -rf /tmp/dfpy/*

cd /app/df-py/
date=`date -dlast-thursday '+%Y-%m-%d'`
now=`date '+%Y-%m-%d'`
nm=`date +%u`

if [ $nm -eq 4 ]
then
        date="$now"
fi
echo $date

/app/df-py/dfpy_docker getrate OCEAN $date $now /app/data
/app/df-py/dfpy_docker getrate ETH $date $now /app/data
/app/df-py/dfpy_docker getrate BNB $date $now /app/data
/app/df-py/dfpy_docker getrate EWT $date $now /app/data
/app/df-py/dfpy_docker getrate MOVR $date $now /app/data
/app/df-py/dfpy_docker getrate MATIC $date $now /app/data

/app/df-py/dfpy_docker volsym $date latest 50 /app/data 1 && 
/app/df-py/dfpy_docker volsym $date latest 50 /app/data 56 && 
/app/df-py/dfpy_docker volsym $date latest 50 /app/data 137 &&
/app/df-py/dfpy_docker volsym $date latest 50 /app/data 246 && 
/app/df-py/dfpy_docker volsym $date latest 50 /app/data 1285 && 

/app/df-py/dfpy_docker vebals  $date latest 50 /app/data 1 &&
/app/df-py/dfpy_docker vebals  $date latest 1 /app/data 1 &&
/app/df-py/dfpy_docker allocations $date latest 50 /app/data 1 
/app/df-py/dfpy_docker allocations $date latest 1 /app/data 1

cp /tmp/dfpy/rate-OCEAN.csv /tmp/dfpy/rate-MOCEAN.csv
sed -i -e 's/MOCEAN/OCEAN/g' /tmp/dfpy/rate-MOCEAN.csv

/app/df-py/dfpy_docker calc /app/data 25000 OCEAN

mv /tmp/dfpy/* ~/.dfcsv/
=========================

nftinfo.sh
=========================
cd /app/df-py/

/app/df-py/dfpy_docker nftinfo /app/data 1
/app/df-py/dfpy_docker nftinfo /app/data 56
/app/df-py/dfpy_docker nftinfo /app/data 137
/app/df-py/dfpy_docker nftinfo /app/data 246
/app/df-py/dfpy_docker nftinfo /app/data 1285
mv /tmp/dfpy/nftinfo* ~/.dfcsv
=========================
```

trentmc:
> Do we just want to test all the functions in 1 complex example like: test_query.py::test_all() and then have many scripts that just use df-py as a base library?

YES. That's exactly what we test_query.py::test_all() is simply a wrapper for many sub-tests, that essentially stand alone. 

The current structure is a very explicit design that Berkay and I arrived based on several problems. 
- There was a time when we had multiple independent unit tests, that each took 3-5 min to populate ganache (create data nfts, DTs, consume, etc). This made runtime super-slow, and unreliable too. 
- The fix was: have a _single_ unified place that populates ganache just _once_, yet maximize independence of tests for sub-modules, That's `test_all()` for the unified place, with the 10+ calls to sub-tests for each of the sub-modules.
- [df-py#429](https://github.com/oceanprotocol/df-py/issues/429) reported this issue. 2-3 months ago I spent a bunch of time to fix this all via several PRs [like this one](https://github.com/oceanprotocol/df-py/commit/bf155ce2fdbfdc44f3f909ef4306cd163cd77556). The end result is what you see.

And of course, test_query::test_all() it is not _all_ the tests for df-py. It's only for tests that need ganache to be well-populated. Obviously we have several other test modules, simply see util/test/test_*.py.

Summary: the design is there for a reason. We arrived at it after past problems & fixes. Please don't break it. Rather, work with it.

----

> 1. What should be library df-py code vs. script code?

Non-dftool script code (bash) == hard to test, and hard to maintain tests. It's also harder than Py to compose to higher levels yet.

Whereas using the pattern of dftool + py sub-modules + py tests, we get all the CLI scripting we want, yet have testability and modularity.

So the answer is: ideal amount of script code is ZERO, except dftool. 

> 2. What can our exploration > ossification process be like. Does it have to be tight? Or can it be loose?

First, let's make sure we are on the same page. Putting into dftool + py sub-modules + py is _not_ ossification. That pattern can change very quickly and readily. In fact, I see it as easier to change, because we can have excellent tests (unlike testing bash), which allows us to be more confident in what changed. 

Which means we don't have to trade off. We get it all, _if_ we are rigorous about it and in sync on the patterns.

----

Summary:
- Structure of test_query.py::test_all() is a solution to a problem of unit test speed & reliability. It maintains low complexity via well-defined sub-tests.
- Any non-dftool script code (bash) is harder to test, to maintain, and to compose to higher levels. 
- Whereas pattern of dftool + py sub-modules + py tests gives us all the CLI scripting we want, and with better testability, maintanability, and composability. (And this makes it _less_ ossifying than bash approach, not more.)

I don't expect that we will move all non-dftool scripts to dftool at once. But we definitely don't want to go the other direction and add more scripts yet. 

That also means for _this_ issue, I _am_ asking for a rewrite of it in the pattern of dftool + py sub-modules + py tests.

idiom-bytes:
```Structure of test_query.py::test_all() is a solution to a problem of unit test speed & reliability. It maintains low complexity via well-defined sub-tests.```

I agree, and it's already testing everything we needed. Now we need code that can orchestrate all of these commands against a live network. That's why the shell scripts were born.

```
Any non-dftool script code (bash) is harder to test, to maintain, and to compose to higher levels.
```

I agree. However, we needed to run all of this against live networks. df-py is pretty generic, it would feel more like we're breaking a pattern by coupling it with networks.

Also df-py is a great cli library, not using bash to explore/play with it seems very restrictive...

```
That also means for this issue, I am asking for a rewrite of it in the pattern of dftool + py sub-modules + py tests.
```
I'll create a separate task to integrate the shell scripts as python sub-modules and close this task.

idiom-bytes:
Hi @trentmc I have created this separate issue in the hopes to close merge this and close this ticket.
https://github.com/oceanprotocol/df-py/issues/492

trentmc:
> As discussed, ticket created to handle next steps. Let's merge this in and then iterate in #492 towards python scripts.

OK. 

Then, approved

trentmc (approved):

jamiehewitt15 (approved):

jamiehewitt15:
@idiom-bytes please pull from main before merging

PR #424, opened 2023-01-10
use OCEAN locked when constraining to 125% APY instead of veOcean

alexcos20:
Fixes # 380

Changes proposed in this PR:

- use locked_amounts instead of veBalance


trentmc:
This won't work.

We need veOCEAN balances (vebals)for all the reward function calculations in the reward function. (Eg it's critical that the amount of reward is a function of how long the OCEAN is locked for.)

The only thing in reward function that will now be a function of just OCEAN locked (not vebals) is to constrain APY.

alexcos20:
Oh, got it, thanks
I will close this PR then

PR #430, opened 2023-01-12
Fix #429: Make tests more reliable & faster

### LINKED ISSUE #429: make test_dftool more reliable & faster
### Goal

Make tests run faster & more reliably

### How

Move the chain-related tests in test_dftool.py into test_query.py. Like what I did for test_end-to-end.py in [#425](https://github.com/oceanprotocol/df-py/issues/425). [Commit](https://github.com/oceanprotocol/df-py/commit/32f15fc27b49627fc7da0e31c9736d5de9e1b63b).

### Background

- test_dftool.py::test_setup() waits for about 30 s at the end of the on-chain work, with the idea that this is "hopefully" long enough for the queries to be ready. However, there's no guarantee, therefore the tests can sporadically fail.
- In https://github.com/oceanprotocol/df-py/commit/bff5520c68ae797ff4d6ef91c864269747369f56 I cleaned up some of test_dftool.py, I moved clearly fast-running tests into a new file test_dftool1.py (and renaming test_dftool.py -> test_dftool2.py).
- That commit was since backed out, because it was buggy

trentmc:
Fixes #429.

### DIFF
--- a/util/test/conftest.py
+++ b/util/test/conftest.py
@@ -1,23 +1,9 @@
 from enforce_typing import enforce_types
 
-import pytest
 import brownie
 
 from util import networkutil
 from util.oceanutil import OCEANtoken, recordDevDeployedContracts
-
-
-@enforce_types
-@pytest.fixture
-def network_setup_and_teardown():
-    networkutil.connect(networkutil.DEV_CHAINID)
-
-    # everyting before the yield is run before the test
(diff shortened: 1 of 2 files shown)


trentmc:
This PR is obsolete. https://github.com/oceanprotocol/df-py/pull/442 is the direction now.

PR #467, opened 2023-01-30
Add dftool `calculate_passive` command

### LINKED ISSUE #443: Add passive rewards amount to vebals.csv
Every time we compute vebals, we can compute amount of passive rewards received by an user for a DF round

Let's add a column, called "amt"  (to keep existing naming, as rewardsInfo is uses amt as well).

And here is how to calculate it:
  - timestamp = beginning of a DF round (for current round: timestamp = current_unix_timestamp / 604800* 604800
  - total_passive_OCEAN= veFeeDistributor.tokens_per_week(timestamp)
  - total_ve_supply = veFeeDistributor.ve_supply(timestamp)
  - OCEAN_per_veOCEAN=total_passive_OCEAN/ total_ve_supply
  - for each user:
  - - user_passive_OCEAN= veFeeDistributor.ve_for_at(user_address,timestamp) * OCEAN_per_veOCEAN

What is "user_passive_OCEAN" ?   amount of Ocean tokens received by a user for passive staking, during that round
Why do this?  So we can store and calculate history of passive rewards

APY calculation in df-web:   easy, since we have user_locked_balance and amount of Ocean received as reward

This introduces another question.  Which veOcean_balance should we use across ve+df ?
 - veFeeDistributor records veOcean_balance per user at the beginning of the round  (see veFeeDistributor.ve_for_at(timestamp,user) )
 - df-py veBals computes an average using 50 snapshots
 
 veFeeDistributor.ve_for_at  >  df-py vebal   (because veOcean balance diminishes during those 50 snapshots)

(This works in tandem with https://github.com/oceanprotocol/df-sql/issues/29), we need to merge them in the same time

trizin:
Fixes #443 

Changes proposed in this PR:

- Added dftool `calculate_passive` command
- Added new query function `queryPassiveRewards`
- Query and dftool command tests
- Fixed black flow to use the version from requirements.txt (in accordance with brownie)

### DIFF
--- a/dftool
+++ b/dftool
@@ -945,7 +945,7 @@ Usage: dftool calculate_passive CHAINID DATE CSV_DIR
     vebals, _, _ = csvs.loadVebalsCsv(CSV_DIR, False)
     addresses = list(vebals.keys())
 
-    balances, rewards = query.queryPassiveRewards(CHAINID, timestamp, addresses)
+    balances, rewards = query.queryPassiveRewards(timestamp, addresses)
 
     # save to csv
     csvs.savePassiveCsv(rewards, balances, CSV_DIR)
--- a/dftool
+++ b/dftool
@@ -945,7 +945,7 @@ Usage: dftool calculate_passive CHAINID AMOUNT DATE CSV_DIR
     vebals, _, _ = csvs.loadVebalsCsv(CSV_DIR, False)
     addresses = list(vebals.keys())
 
-    balances, rewards = query._queryPassiveRewards(CHAINID, timestamp, addresses)
+    balances, rewards = query.queryPassiveRewards(CHAINID, timestamp, addresses)
 
     # save to csv
     csvs.savePassiveCsv(rewards, balances, CSV_DIR)
--- a/dftool
+++ b/dftool
@@ -914,7 +914,7 @@ Usage: dftool dispense_passive CHAINID AMOUNT
 def do_calculate_passive():
     HELP = f"""Calculate passive rewards.
 
-Usage: dftool calculate_passive CHAINID AMOUNT DATE CSV_DIR
+Usage: dftool calculate_passive CHAINID DATE CSV_DIR
     CHAINID -- {CHAINID_EXAMPLES}
     DATE -- date in format YYYY-MM-DD
     CSV_DIR -- output dir for passive-CHAINID.csv
--- a/dftool
+++ b/dftool
@@ -919,14 +919,14 @@ Usage: dftool calculate_passive CHAINID DATE CSV_DIR
     DATE -- date in format YYYY-MM-DD
     CSV_DIR -- output dir for passive-CHAINID.csv
 """
-    if len(sys.argv) not in [6]:
+    if len(sys.argv) not in [5]:
         print(HELP)
         sys.exit(0)
 
     CHAINID = int(sys.argv[2])
     networkutil.connect(CHAINID)
-    DATE = sys.argv[4]
-    CSV_DIR = sys.argv[5]
+    DATE = sys.argv[3]
+    CSV_DIR = sys.argv[4]
     timestamp = int(timestrToTimestamp(DATE))
     S_PER_WEEK = 7 * 86400
     timestamp = timestamp // S_PER_WEEK * S_PER_WEEK
--- a/dftool
+++ b/dftool
@@ -27,7 +27,8 @@ from util.oceanutil import (
     veAllocate,
     FeeDistributor,
 )
-from util.blocktime import getfinBlock
+
+from util.blocktime import getfinBlock, timestrToTimestamp
 from util.multisig import send_multisig_tx
 
 brownie.network.web3.middleware_onion.inject(geth_poa_middleware, layer=0)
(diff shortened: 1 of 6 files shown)


dftool
```diff
@@ -911,6 +909,48 @@ Usage: dftool dispense_passive CHAINID AMOUNT
     print("Dispensed passive rewards")
 
 
+# ========================================================================
+@enforce_types
+def do_calculate_passive():
+    HELP = f"""Calculate passive rewards.
+
+Usage: dftool calculate_passive CHAINID AMOUNT DATE CSV_DIR
```
idiom-bytes, line 917: :heart_eyes:

util/query.py
```diff
@@ -492,6 +492,50 @@ def _queryNftvolumes(
...
+    total_rewards_float = fromBase18(total_rewards)
+
+    if sum_bals_float == 0:
+        return balances, rewards
+
+    for addr in addresses:
+        balance = fee_distributor.ve_for_at(addr, timestamp)
+        balance_float = fromBase18(balance)
+        balances[addr] = balance_float
+        rewards[addr] = total_rewards_float * balance_float / sum_bals_float
```
idiom-bytes, line 534: Fantastic work.

dftool
```diff
@@ -911,6 +909,48 @@ Usage: dftool dispense_passive CHAINID AMOUNT
...
+    _exitIfFileExists(passive_fname)
+
+    # get addresses
+    vebals, _, _ = csvs.loadVebalsCsv(CSV_DIR, False)
+    addresses = list(vebals.keys())
+
+    balances, rewards = query.queryPassiveRewards(CHAINID, timestamp, addresses)
+
+    # save to csv
+    csvs.savePassiveCsv(rewards, balances, CSV_DIR)
```
idiom-bytes, line 951: :+1: 

@alexcos20 are you using this data to feed df-sql like the rest of the pipeline?

util/csvs.py
```diff
@@ -175,6 +175,43 @@ def vebalsCsvFilename(csv_dir: str, sampled=True) -> str:
...
+      rewards -- dict of [LP_addr] : reward
+      balances -- dict of [LP_addr] : balance
+      csv_dir -- directory that holds csv files
+    """
+    assert os.path.exists(csv_dir), csv_dir
+    csv_file = passiveCsvFilename(csv_dir)
+    assert not os.path.exists(csv_file), csv_file
+    with open(csv_file, "w") as f:
+        writer = csv.writer(f)
+        row = ["LP_addr", "balance", "reward"]
```
idiom-bytes, line 198: :raised_hands:

util/test/test_query.py
```diff
@@ -735,6 +736,40 @@ def test_populateNftAssetNames():
...
+    fee_distributor.checkpoint_total_supply({"from": brownie.accounts[0]})
+    fee_distributor.checkpoint_token({"from": brownie.accounts[0]})
+    chain.sleep(S_PER_WEEK * 2)
+    chain.mine()
+    fee_distributor.checkpoint_total_supply({"from": brownie.accounts[0]})
+    fee_distributor.checkpoint_token({"from": brownie.accounts[0]})
+    balances, rewards = query.queryPassiveRewards(CHAINID, chain.time(), addresses)
+    print(balances)
+    print(rewards)
+    assert len(balances) == 90
```
idiom-bytes, line 770: I follow the logic of the bootstrap and how it's distributing passive rewards, to then query the balance via subgraph.

But the test is checking for 90 balances.
Will this test pass across all of our workflows?

This seems to be configured against the stress test.

idiom-bytes (commented):
Left some comments.

util/test/test_query.py
trizin: Test is not ready yet :face_in_clouds: I added the assert statement to fail the test flow

dftool
alexcos20: yeap

jamiehewitt15:
By the way, on a general point, we should update the CI to run a code coverage check so we can be sure that each PR has full test coverage before merging.

util/test/test_query.py
```diff
@@ -735,6 +736,32 @@ def test_populateNftAssetNames():
...
+    chain.mine()
+    fee_distributor.checkpoint_token({"from": brownie.accounts[0]})
+    fee_distributor.checkpoint_total_supply({"from": brownie.accounts[0]})
+    balances, rewards = query.queryPassiveRewards(CHAINID, chain.time(), addresses)
+    print(balances)
+    print(rewards)
+    assert balances[0] == balances[1]
+    assert rewards[0] == rewards[1]
+    assert rewards[0] > 0
+
```
jamiehewitt15, line 763: I'd suggest that we should also have a test that reads the rewards and balances out of the local csv, so that we can be sure the the csv has successfully been saved in the right location with the right name and headings.

util/test/test_query.py
```diff
@@ -735,6 +736,32 @@ def test_populateNftAssetNames():
...
+    OCEAN = oceanutil.OCEANtoken()
+
+    OCEAN.transfer(
+        fee_distributor.address,
+        toBase18(1000.0),
+        {"from": brownie.accounts[0]},
+    )
+    fee_distributor.checkpoint_token({"from": brownie.accounts[0]})
+    fee_distributor.checkpoint_total_supply({"from": brownie.accounts[0]})
+    chain.sleep(S_PER_WEEK * 3)
```
alexcos20, line 753:
```suggestion
   for i in range(0, 3):
       chain.sleep(S_PER_WEEK)
       chain.mine()
       fee_distributor.checkpoint_token({"from": brownie.accounts[0]})
       fee_distributor.checkpoint_total_supply({"from": brownie.accounts[0]})
```
then do balances, rewards = query.queryPassiveRewards(CHAINID, chain.time(), addresses)

also make sure that lock_time for those addresses is long enough

alexcos20 (approved):

jamiehewitt15 (approved):
This looks good to me

PR #490, opened 2023-02-28
Finish #411: removed references to "dftool dispense" and "query".

### LINKED ISSUE #411: "dftool query" and "dftool dispense" need more specific names
### Background / motivation

In dftool, CLI includes:
```text
  dftool query ST FIN NSAMP CSV_DIR CHAINID - query chain, get (symbols, vols)
  dftool allocations ST FIN NSAMP CSV_DIR CHAINID
  dftool vebals ST FIN NSAMP CSV_DIR CHAINID
...
  dftool dispense CSV_DIR [CHAINID] [DFREWARDS_ADDR] [TOKEN_ADDR] - from rewards, dispense funds
  dftool dispense_passive CHAINID AMOUNT
```

The first 3 commands are about querying info from the chain
- symbols and volumes
- allocations
- vebals

The last two commands are about dispensing funds
- active
- passive

The problems:
- The first query command (`dftool query`) doesn't distinguish _what_ it's querying and returning. Same for the function it calls (`dftool::do_query()`).
- The first dispense command (`dftool dispense`) doesn't distinguish whether it's passive or active. Same for the function it calls (`dftool::do_dispense()`).

### TODO for this issue

- [x] Update ["flows" GSlides](https://docs.google.com/presentation/d/11Nu_emSgw3iYZKDcmnvZC_aOM00MDyYTtTWQP4PYox4/edit#slide=id.g148220d8488_0_0) 
- [x] (a) rename do_query() to do_volsim(); (b) Rename do_dispense() to do_dispense_active()
- [x] (a) add "dftool volsim" which calls do_volsim(); keep "dftool query" at this point. (b) add "dftool dispense_active" which calls do_dispense_active(); keep "dftool dispense" at this point.
- [x] in df-sql, df-webapp, github actions, or anywhere else that uses "dftool query", replace with "dftool volsim". Same for "dftool dispense" --> "dftool dispense_active"
- [x] once all of (3) is clearly working, then deprecate "dftool query" and "dftool dispense"

idiom-bytes:
Finish #411, namely:
>  once all of (3) is clearly working, then deprecate "dftool query" and "dftool dispense"

### DIFF
--- a/util/test/test_dftool.py
+++ b/util/test/test_dftool.py
@@ -118,11 +118,7 @@ def test_manyrandom():
 def test_noarg_commands():
     # Test commands that have no args. They're usually help commands;
     # sometimes they do the main work (eg compile).
-    argv1s = [
-        "",
-        "compile",
-        "newacct"
-    ]
+    argv1s = ["", "compile", "newacct"]
     for argv1 in argv1s:
         cmd = f"./dftool {argv1}"
 
--- a/util/test/test_dftool.py
+++ b/util/test/test_dftool.py
@@ -118,8 +118,24 @@ def test_manyrandom():
 def test_noarg_commands():
     # Test commands that have no args. They're usually help commands;
     # sometimes they do the main work (eg compile).
-    argv1s = ["", "compile", "newacct"]
+    argv1s = [
+        "",
+        "volsym",
+        "getrate",
+        "calc",
+        "dispense_active",
+        "querymany",
+        "compile",
+        "manyrandom",
+        "newdfrewards",
+        "mine",
+        "newacct",
+        "newtoken",
+        "acctinfo",
+        "chaininfo",
+    ]
     for argv1 in argv1s:
+        print(f"Test dftool {argv1}")
         cmd = f"./dftool {argv1}"
 
         output_s = ""
@@ -131,7 +147,7 @@ def test_noarg_commands():
 
         return_code = proc.wait()
         # bad commands - such as querymany - will still return 0 and do not fail
-        assert return_code == 0, print(f"'dftool {argv1}' failed. \n{output_s}")
+        assert return_code == 0, f"'dftool {argv1}' failed. \n{output_s}"
 
 
 @enforce_types
--- a/util/test/test_dftool.py
+++ b/util/test/test_dftool.py
@@ -120,22 +120,10 @@ def test_noarg_commands():
     # sometimes they do the main work (eg compile).
     argv1s = [
         "",
-        "volsym",
-        "getrate",
-        "calc",
-        "dispense_active",
-        "querymany",
         "compile",
-        "manyrandom",
-        "newdfrewards",
-        "mine",
-        "newacct",
-        "newtoken",
-        "acctinfo",
-        "chaininfo",
+        "newacct"
     ]
     for argv1 in argv1s:
-        print(f"Test dftool {argv1}")
         cmd = f"./dftool {argv1}"
 
         output_s = ""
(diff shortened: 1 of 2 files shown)


idiom-bytes:
Moved dftool functionality and other tests that need to be created onto other ticket. 
https://github.com/oceanprotocol/df-py/issues/491

util/test/test_dftool.py
```diff
@@ -118,26 +118,8 @@ def test_manyrandom():
...
-        "querymany",
-        "compile",
-        "manyrandom",
-        "newdfrewards",
-        "mine",
-        "newacct",
-        "newtoken",
-        "acctinfo",
-        "chaininfo",
-    ]
```
trentmc, line 138: This should not be removed. It's really important to keep, since it gives at least super-basic tests to many dftool CLI tools that don't otherwise tests.

Instead, just simply remove the lines for "query" and "dispense"

util/test/test_dftool.py
```diff
@@ -118,26 +118,8 @@ def test_manyrandom():
...
+    argv1s = ["", "compile", "newacct"]
```
trentmc, line 121: In line with previous comment: do _not_ add this line.

util/test/test_dftool.py
```diff
@@ -118,26 +118,8 @@ def test_manyrandom():
...
     for argv1 in argv1s:
-        print(f"Test dftool {argv1}")
```
trentmc, line 140: Do _not_ remove this. It's there because it really helps debugging.

util/test/test_dftool.py
```diff
@@ -148,7 +130,8 @@ def test_noarg_commands():
                 output_s += proc.stdout.readline().decode("ascii")
 
         return_code = proc.wait()
-        assert return_code == 0, f"'dftool {argv1}' failed. \n{output_s}"
+        # bad commands - such as querymany - will still return 0 and do not fail
+        assert return_code == 0, print(f"'dftool {argv1}' failed. \n{output_s}")
```
trentmc, line 134: This "change" is wrong, because it does not follow proper `assert` syntax. It was fine before, don't change it.

trentmc (changes requested):

util/test/test_dftool.py
trentmc: More info:

In slack @idiom-bytes had mentioned that they were removed "because they were not doing anything". This is incorrect. Ask yourself "why might these possibly be here?" I'll make life simple and give the answer: `dftool X` with no extra arguments goes to that method, and gives the _help_ info for that command. And that's what this is testing. 

Extra benefit: many dftool commands _only_ have this as a test, so this is the only pushes down into that dftool command's `do_X()` function.

idiom-bytes: Yup, as discussed in our meeting. Thank you.

util/test/test_dftool.py
idiom-bytes: I see, thank you.

idiom-bytes: I have kept the comment for prosperity of understanding what's going on inside the subprocess (i.e. we know some tests fail but that's the test), and reverted the print statement.

util/test/test_dftool.py
idiom-bytes: Reverted

util/test/test_dftool.py
idiom-bytes: Reverted to the full list and removed query/dispense.

trentmc (approved):
Nice and clean. Does exactly what the PR says it should. Which is exactly the github issue.

Which makes it easy for me to approve! 👍👍

jamiehewitt15 (approved):
Looks good to me

PR #507, opened 2023-03-15
Expanding dftool functionality

### LINKED ISSUE #270: Expand dftool interface to make debugging ve functionality easier [one last thing: close off final PR]
Expanded commands to:
- create lock
- increment amount
- get lock end time

Currently veOcean is behaving a bit funky: reverting calls suggesting user already has lock, and estimating astronomical gas costs. Implemented these to make debugging & testing easier.

idiom-bytes:
Fixes #270 

Changes proposed in this PR:
- Add more helper function that talks to veOCEAN and other contracts
- get_balance_veocean, get_lock_end_veocean, increase_unlock_time_veocean, increase_lock_amount_veocean

### DIFF
--- a/dftool
+++ b/dftool
@@ -27,6 +27,11 @@ from util.oceanutil import (
     recordDeployedContracts,
     OCEANtoken,
     veAllocate,
+    create_lock_veocean,
+    increase_amount_veocean,
+    increase_unlock_time_veocean,
+    get_balance_veocean,
+    get_lock_end_veocean,
     FeeDistributor,
 )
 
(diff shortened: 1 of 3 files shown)


trentmc (approved):

jamiehewitt15 (approved):

idiom-bytes:
Noticed a couple things:
- adding tests
- adding command withdraw_veocean
- improving documentation

jamiehewitt15:
There are also conflicts that need to be resolved

jamiehewitt15:
There are still conflicts that need to be resolved here. 

Is this still actively being worked on? There hasn't been any activity since 15th March

idiom-bytes:
I revised the tests quite a bit because they weren't really implemented properly. The key-thing here is using alice taking over the os.env['DFTOOL_KEY'] so every test is it's own instance and won't impact the rest.

I played around with it in the cli and it felt right.

idiom-bytes:
Can we follow up tomorrow @trizin to see if we can debug this? I know you kicked off another build but it seems like it failed.

idiom-bytes:
This is very close to shipping but let's focus on what matters.
We're going to shelve this for now and come back to it when it settles.

PR #513, opened 2023-03-19
Use fixedRateExchangeSwaps

### LINKED ISSUE #512: Fix ghost consume
Fix ghost consume

trizin:
Fixes #512

Changes proposed in this PR:

- Use "fixedRateExchangeSwaps" events instead of "order" events

For ref: [opf slack thread](https://oceanprotocol.slack.com/archives/C026BNFH6HK/p1679275743495979?thread_ts=1679157220.832099&cid=C026BNFH6HK)

### DIFF
--- a/util/test/test_query.py
+++ b/util/test/test_query.py
@@ -874,6 +874,14 @@ def _test_queryPassiveRewards(addresses):
         sim_epoch()
 
 
+@enforce_types
+def test_filter_by_max_volume():
+    nftvols = {"a": {"b": 1000}}
+    swapvols = {"a": {"b": 100}}
+    filteredvols = query._filterbyMaxVolume(nftvols, swapvols)
+    assert filteredvols["a"]["b"] == 100
+
+
 # ===========================================================================
 # support functions
 


util/query.py
```diff
@@ -446,27 +447,34 @@ def _queryVolsOwners(
         result = submitQuery(query, chainID)
         if "errors" in result:
             raise AssertionError(result)
-        new_orders = result["data"]["orders"]
+        new_orders = result["data"]["fixedRateExchangeSwaps"]
```
trentmc, line 450: In DF, we want to account for 
1. selling of the data asset
2. consuming of the token
3. And ideally: include free assets where price = gas cost)

Before the PR, it did all of (1)(2)(3) albeit with a potential error. 

What this PR does: switch "orders" (the first step of a _consume_ - 2) with "fixedRateExchangeSwap" (a sale - 1).

What's the coverage of this PR?
- has (1)
- doesn't have (2)
- doesn't have (3). 

If needed we could skip (3) temporarily, I guess. 

But should we be skipping (2) ? I'm dubious.

trentmc (approved):

PR #537, opened 2023-05-01
(try 1) Fix #539: Port pej into df-py

### LINKED ISSUE #539: Port predict-eth-judge into df-py repo, towards Challenge DF
### TODOs

- [ ] Create "Challenge DF" branch in df-py repo. [old PR](https://github.com/oceanprotocol/df-py/pull/537) [new PR](https://github.com/oceanprotocol/df-py/pull/559)
- [ ] Port code from predict-eth-judge repo into here. This will compute the 3 winners, and how much $ each should get. _Don't_ hook into dftool in this issue.

Note: the above is all that's needed for _this_ issue. [#540](https://github.com/oceanprotocol/df-py/issues/540) will take over from there.

calina-c:
Fix #539


alexcos20:
why use something [external and mock it](https://github.com/oceanprotocol/df-py/pull/537/files#diff-089563ddf50284768e89d9e879812088b39ab5320ed646979c0d25e173c687c4R48-R55)  when you could just use a simple subgraph query like [this](https://v4.subgraph.mumbai.oceanprotocol.com/subgraphs/name/oceanprotocol/ocean-subgraph/graphql?query=query%7B%0A++nftTransferHistories%28where%3A+%7Btimestamp_gt%3A+0%2C+timestamp_lt%3A1768629481%7D+skip%3A0+first%3A1000%29%7B%0A++++nft+%7B%0A++++++id%0A++++++owner+%7B%0A++++++++id%0A++++++%7D%0A++++%7D%0A++++oldOwner+%7B%0A++++++id%0A++++%7D%0A++++newOwner+%7B%0A++++++id%0A++++%7D%0A++++timestamp%0A++++txId%0A++%7D%0A%7D) (replace timestamps, do a loop in case we have > 1000 results, filter with newOwner=='0xa54abd42b11b7c97538cad7c6a2820419ddf703e' , etc)  and do a proper integration test in barge?  (with actual publish, transfer, etc)

trentmc:
> why not just use a simple subgraph query ?

This is a good idea.

I want to get _this_ PR merged in ASAP though. So that I can incorporate it into the overall "3-substreams" flow for df-main.

Therefore let's defer this suggestion to follow-up work. I just created a new issue [df-py#543](https://github.com/oceanprotocol/df-py/issues/543) for exactly that.

trentmc:
FYI I reported a CI error that this PR gets:
https://github.com/oceanprotocol/df-py/issues/547

util/judge.py
```diff
@@ -0,0 +1,188 @@
+import os
+import sys
+from datetime import datetime as dt
+from datetime import timedelta
+from typing import List
+
+import ccxt
+import requests
+from enforce_typing import enforce_types
+from ocean_lib.models.data_nft import DataNFT
```
trentmc, line 10: df-py already has its way of accessing ocean with a "thin wrapper" of the contracts, via `util/oceanutil.py`. By design, it avoids importing ocean.py because of the large additional complexity.

Please leverage it here. Avoid adding in ocean.py.

util/judge.py
```diff
@@ -0,0 +1,188 @@
...
+from ocean_lib.ocean import crypto
```
trentmc, line 11: Rather than relying on ocean.py here, copy & paste the relevant function from ocean.py's [crypto.py](https://github.com/oceanprotocol/ocean.py/blob/main/ocean_lib/ocean/crypto.py) to here.

trentmc (changes requested):

util/judge.py
calina-c: ok! ocean.py is a dependency of predict-eth-helpers, though. I will also have to copy the helpers I'm using.

trentmc: Yes, do copy. We don't want that dependency either.

dftool
```diff
@@ -4,6 +4,7 @@ import datetime
 import functools
 import os
 import sys
+from calendar import WEDNESDAY
```
trentmc, line 7: @calina-c you can remove the changes in dftool. I will handle them as part of [issue#540](https://github.com/oceanprotocol/df-py/issues/540) // [PR#554](https://github.com/oceanprotocol/df-py/pull/554).

Issue #434, opened 2023-01-12
Reconfigure flow.yml and other bootstrappres to use infura rpc + WEB3_INFURA_PROJECT_ID


idiom-bytes:
### Problem
We can't rely on public RPCs.
Active Rewards continued to fail until we found a public RPC that was table.

The flow.yml brownie bootstrapper fails in these situations
```
brownie networks add moonbase moonbase host=https://rpc.api.moonbase.moonbeam.network chainid=1287
brownie networks add mumbai mumbai host=https://matic-mumbai.chainstacklabs.com chainid=80001
brownie networks add energyweb energyweb host=https://rpc.energyweb.org chainid=246
brownie networks add polygon polygon host=https://matic-mainnet.chainstacklabs.com chainid=137
brownie networks add bsc bsc host=https://bsc-dataseed1.binance.org chainid=56
brownie networks add moonriver moonriver host=https://rpc.api.moonriver.moonbeam.network chainid=1285
```

### Solution
1. Modify our yml files, bootstrappers, and documentation, to use infura.
2. Make sure that the `brownie network add` command pipes the cout to null ` >> /dev/null`
3. Make sure that the brownie/request logic, is able to talk to the infura RPC provider. Right now, updating this configuration causes an HTTP request error to be thrown. I'll share this below. 

### Review:
Step 1 - Update bootstrapper to use infura RPCs for each network (wherever available) and to dereference the WEB3_INFURA_PROJECT_ID env_var. 
```
...
brownie networks add polygon polygon host=https://polygon-mainnet.infura.io/v3/${WEB3_INFURA_PROJECT_ID} chainid=137
...
```

Step 2 - Fix the brownie HTTP request to work w/ infura. If we just modify the above, the following error is triggered. So, please update whatever needs to happen internally for infura to work w/ brownie.
```
Traceback (most recent call last):
  File "./dftool", line 1053, in <module>
    do_main()
  File "/home/anon/df-py/venv/lib/python3.8/site-packages/enforce_typing/decorator.py", line 29, in wrapper
    return func(*args, **kwargs)
  File "./dftool", line 1001, in do_main
    do_volsym()
  File "/home/anon/df-py/venv/lib/python3.8/site-packages/enforce_typing/decorator.py", line 29, in wrapper
    return func(*args, **kwargs)
  File "./dftool", line 144, in do_volsym
    networkutil.connect(CHAINID)
[...]

PR #586, opened 2023-05-16
Intermediary work for Infura keys.

calina-c:
Work on #434.


calina-c:
@trizin I know you recently changed the polygon RPC. If it is reliable, should I change it to Infura?

PR #605, opened 2023-05-22
Dockerfile updates after introducing setup.py

calina-c:
The docker builds (also recently introduced) started failing after introducing setup.py and versioning. This should fix the builds.


calina-c:
Before merging, I need to do a bumpversion as well. For now, let's see if the fixes work :)

Issue #505, opened 2023-03-15
Fix dftool commands to use snake_case


idiom-bytes:
### Problem:
- [x] There are bunch of inconsitencies in dftool command names. Fix them so they are using snake_case and consistent.
- [ ] VM, GHA, and other flows could be impacted by this. We do not want to break things at the same time.
- [ ] Let's hold on this one (i.e. think of a point releases perhaps)

### DoD:
- [ ] Figure out when/how to proceed.
- [ ] Use this format from here on out.

trentmc:
Redundant with #618 which is more general, so closing this

PR #611, opened 2023-05-24
Add predictoor models, queries and dftool predictoor data command

### LINKED ISSUE #610: Write `dftool predictoor_data`
Write `dftool predictoor_data`. There's already a stub. Use similar pattern to challenge_data.

trizin:
Fixes #610

Changes proposed in this PR:

- Added predictoor models
- Added predictoor query function
- Added predictoor data csv functions
- Updated dftool predictoor data command
- Fix load predictoor data csv function
- Tests for predictoor models
- Tests for predictoor query
- Tests for new csv functions
- Removed Sidestaking contract as it was deprecated
- Only query whitelisted dt3 templates, whitelisted = deployed by opf
- Test new dftool command.
- Test the predictoor query function without subgraph mock to validate the subgraph query (skipping, tested locally)

TODO:
- ~Use predictoor barge~ **Predictoor barge continuously mining new blocks and progressing the chain's time. This new behaviour breaks the current tests.**

### DIFF
--- a/util/csvs.py
+++ b/util/csvs.py
@@ -200,7 +200,7 @@ def saveChallengeDataCsv(challenge_data: tuple, csv_dir: str):
         writer = csv.writer(f)
         row = ["from_addr", "nft_addr", "nmse"]
         writer.writerow(row)
-        for from_addr, nft_addr, nmse in zip(from_addrs, nft_addrs, nmses):
+        for (from_addr, nft_addr, nmse) in zip(from_addrs, nft_addrs, nmses):
             assertIsEthAddr(from_addr)
             assertIsEthAddr(nft_addr)
             row = [


util/predictoor/models.py
```diff
@@ -0,0 +1,47 @@
...
+
+class Prediction:
+    @enforce_types
+    def __init__(self, block: int, payout: float, contract_addr: str):
+        self.block = block
+        self.payout = payout
+        self.contract_addr = contract_addr
+
+    @property
+    def correct(self) -> bool:
```
calina-c, line 13: I would suggest `is_correct` to emphasize that we are not correcting something, we are checking if something is correct

util/predictoor/query.py
```diff
@@ -0,0 +1,80 @@
+from typing import Dict
+
+from util.graphutil import submitQuery
+from util.predictoor.models import Predictoor, Prediction
+
+
+def queryPredictoors(st_block: int, end_block: int, chainID: int):
```
calina-c, line 7: please use snake case for functions

util/predictoor/query.py
```diff
@@ -0,0 +1,80 @@
...
+        result = submitQuery(query, chainID)
+        if "error" in result:
+            raise AssertionError(result)
+
+        predictions = result["data"]["predictPredictions"]
+        if len(predictions) == 0:
+            break
+
+        for prediction in predictions:
+            predictoor_addr = prediction["user"]["id"]
```
calina-c, line 63: how about a static function on Prediction called `from_query_result`? It could take in a dict and return the object

calina-c (commented):
I think there is potential to abstract the CSV read/write part separately. We seem to have a lot of functions that do three things:

- generate OR explore the data
- format the data 
- save them to csv OR read a csv

Could we separate the concerns here and have some helpers for reading/writing dictionary data to/from the csvs, while only concerning ourselves with generating/exploring the data in these particular functions?

This can be a separate issue, of course. But I think it is warranted by this point, since we seem to keep adding these types of functions.

trentmc:
> I think there is potential to abstract the CSV read/write part separately
> Could we separate the concerns here and have some helpers for reading/writing dictionary data to/from the csvs, 

This is exactly what csvs.py does.

This pattern has served us extremely well so far. Please let's not change it here, or elsewhere.

Once a given type of data has functionality in csvs.py - with appropriate testing - it is extremely easy to maintain.

util/predictoor/query.py
calina-c: yes, true, I want to discuss this further because it seems we have a mixture of patterns. Previously we did not have a lot of classes, so it was a convention-like thing. But as this repo/project grows we might think about implementing casing standards.

util/predictoor/query.py
```diff
@@ -0,0 +1,79 @@
...
+    """
+    predictoors: Dict[str, Predictoor] = {}
+
+    chunk_size = 1000
+    offset = 0
+
+    while True:
+        # pylint: disable=line-too-long
+        query = """
+        {
```
calina-c, line 34: I would suggest moving all the raw queries in the project to a separate file or module. However, this seems low priority. I suggest creating a ticket if you agree. Otherwise please mark this comment as resolved.

calina-c (commented):
Added some questions and proposals. This looks like amazing work 🎉

util/predictoor/query.py
trentmc: > implementing casing standards as the project grows is beneficial for consistency

I'm ok if we create a new github issue for this. In df-py, I was using casing standards that no longer the standard in python, we really should switch to the proper ones (snake case for function names, etc). 

It could be something for Calina to do, to help learn the codebase.

> suggest moving all the raw queries in the project to a separate file or module

OK, though that was kinda the original point of query.py, hence its name. 

Better: 
- move most challenge DF-related queries into df-py/util/challenge/query.py. And other new modules from challenge (like I've already started).
- similarly, move predictoor DF-related queries into df-py/util/predictoor/query.py
- and to wrap it up, move volume DF-related queries into df-py/util/volume/query.py. And move other volume-specific DF modules there (many!)

trentmc: Re the new directories:
- for _this_ issue just move the new predictoor DF-related queries into df-util/predictoor/query.py
- and create a new issue for the rest

calina-c: > Re the new directories:
> 
> * for _this_ issue just move the new predictoor DF-related queries into df-util/predictoor/query.py
> * and create a new issue for the rest

I created https://github.com/oceanprotocol/df-py/issues/617 for moving the queries.
Should I also create a ticket for casing? Or does https://github.com/oceanprotocol/df-py/issues/505 suffice for now?

trentmc: > Should I also create a ticket for casing? Or does #505 suffice for now?

#505 is just for dftool. Please create a separate issue.

calina-c: Created #618. I will wait for:

> for this issue just move the new predictoor DF-related queries into df-util/predictoor/query.py

in order to mark this issue as resolved.

util/dftool_module.py
```diff
@@ -19,8 +19,9 @@
     networkutil,
     query,
 )
+from util.predictoor.query import queryPredictoors
```
trentmc, line 22: Remove. The import to `query` above is enough. Then below, just use it as `query.queryPredictors`. 

Why: the way it's currently written, now you have to write `queryPredictoors` twice. Harder to maintain 2 copies of every query function name in this module, versus a single `import query`. Calina and I discussed. At the very least, we need to maintain consistency with the current way.

util/dftool_module.py
```diff
@@ -467,50 +468,55 @@ def do_challenge_data():
...
-    chain = brownie.network.chain  # pylint: disable=unused-variable
-    recordDeployedContracts(ADDRESS_FILE)
+    chain = brownie.network.chain
+
+    st_block, fin_block = getstfinBlocks(chain, ST, FIN)
 
     # main work
     predictoor_data = retryFunction(
-        query.queryPredictoor_Data,
+        queryPredictoors,
```
trentmc, line 512: Change to `query.queryPredictoors` like mentioned above.

util/test/predictoor/test_models.py
```diff
@@ -0,0 +1,62 @@
...
+    assert prediction.slot == 123
+    assert prediction.payout == 1.23
+    assert prediction.contract_addr == "0x1"
+
+
+def test_prediction_is_correct():
+    prediction = Prediction(123, 1.23, "0x1")
+    assert prediction.is_correct is True
+    prediction = Prediction(123, 0.0, "0x1")
+    assert prediction.is_correct is False
```
trentmc, line 17: Change to `assert not prediction.is_correct`

util/test/predictoor/test_models.py
```diff
@@ -0,0 +1,62 @@
...
+def test_prediction_init():
+    prediction = Prediction(123, 1.23, "0x1")
+    assert prediction.slot == 123
+    assert prediction.payout == 1.23
+    assert prediction.contract_addr == "0x1"
+
+
+def test_prediction_is_correct():
+    prediction = Prediction(123, 1.23, "0x1")
+    assert prediction.is_correct is True
```
trentmc, line 15: Change to `assert prediction.is_correct`

util/test/test_dftool_ganache.py
```diff
@@ -50,6 +51,21 @@ def test_calc(tmp_path):
...
+    FIN = "latest"
+
+    cmd = f"./dftool predictoor_data {CSV_DIR} {ST} {FIN} {CHAINID}"
+    os.system(cmd)
+
+    # test result
+    predictoor_data_csv = csvs.predictoorDataFilename(CSV_DIR, CHAINID)
+    assert os.path.exists(predictoor_data_csv)
+
+
```
trentmc, line 68: You can improve this test:
- Don't aim to test all the functionality end-to-end. Including: don't test subgraph here, you can do that in the specific dedicated unit test.
- Rather, change dftool_module.py::do_predictoor_data to create fake data when it detects that it's in a test. "Design for test (DFT)"; it's a common technique. Then you can also test whether it handles arguments & optional arguments well too.
- I do this already for testing challenge data. Check out [test_dftool_challenge.py ](https://github.com/oceanprotocol/df-py/blob/main/util/challenge/test/test_dftool_challenge.py) and [dftool_module::do_challenge_data](https://github.com/oceanprotocol/df-py/blob/fb0b19d3971c201fdbd815328593368422397979/util/dftool_module.py#L399). See:
```python
    if judge.DFTOOL_TEST_FAKE_CSVDIR in CSV_DIR:
        challenge_data = judge.DFTOOL_TEST_FAKE_CHALLENGE_DATA

    else:  # main path
...
```

util/test/predictoor/test_models.py
```diff
@@ -0,0 +1,62 @@
+import pytest
```
trentmc, line 1: We already discussed this. The directory structure needs to change. This test & siblings need to move from `util/test/predictoor/test_*.py` to `util/predictoor/test_*.py`. It's more maintainable to have tests at leaf nodes, versus replicate a hierarchy of directories.

trentmc (changes requested):

util/constants.py
```diff
@@ -33,3 +35,7 @@
 
 # filled in by oceanutil.py
 CONTRACTS: dict = {}  # [chainID][contract_label] : contract_object
+
+# predictoor
+MIN_PREDICTIONS = 1818
+DEPLOYER_ADDRS: List[str] = []
```
calina-c, line 41: I don't quite understand this pattern. I see we have something similar for CONTRACTS but these are not constants. Are they supposed to be global variables? Configs?

util/csvs.py
```diff
@@ -254,6 +255,78 @@ def challengeDataCsvFilename(csv_dir: str) -> str:
     return os.path.join(csv_dir, f)
 
 
+# ========================================================================
+# predictoor csv
+
+
+@enforce_types
```
calina-c, line 262: please add these to predictoor/csvs.py to match up with the modularising task and avoid conflicts. Also, please move predictoor-related tests under util/predictoor/test instead of util/test/predictoor (at least for now, it is what Trent recommended).

util/dftool_module.py
calina-c: I would like to mention that we discussed, but I lean towards how Berkay did it here. We did not reach any conclusion. However, we can keep consistency for now and "argue" later :)

util/predictoor/query.py
```diff
@@ -0,0 +1,97 @@
...
+        if "data" not in result:
+            raise AssertionError(result)
+
+        predictions = result["data"]["predictPredictions"]
+        if len(predictions) == 0:
+            break
+
+        for prediction_dict in predictions:
+            owner = prediction_dict["slot"]["predictContract"]["token"]["nft"]["owner"]
+            if chainID != DEV_CHAINID:
```
calina-c, line 81: I think you can combine these ifs with an or.
Also, I would suggest handling these assertions in from_query_result. I.e. try creating the object (no adding it to the list yet). Catch any exceptions from its creation and "continue" if an Exception is caught. In this way you avoid processing these dicts in two places.

util/test/predictoor/test_models.py
calina-c: Yep sorry, I didn't see Trent's comment before commenting the same thing a few lines above. But Berkay also didn't know about this when he started the PR :)

util/test/test_dftool_ganache.py
calina-c: The thing is, mocks don't work that well when coupled with command line. So they don't record coverage and we can't patch command lines. Furthermore, this test is exactly the one that fails when ran out of order, during my experiments with modularising. Which is me and Berkay discussed a strategy to improve code coverage (see #624) where we could have a 2-pronged approach:

- test the functionality of `do_<something>` functions through pytest
- test the calling of `do_<something>` functions through df_module CLI separately

this way we can use patches more freely and make sure the correct coverage is picked up and mocks work when needed.

calina-c (changes requested):

util/dftool_module.py
trentmc: In this case I guess we can't use
from util.predictoor import query 

Because that would be a different query module 

So I guess leave it as-is here

util/test/predictoor/test_models.py
trentmc: Yes

util/dftool_module.py
```diff
@@ -20,7 +20,7 @@
     query,
 )
 from util.predictoor import csvs as predictoor_csvs
-from util.predictoor.query import queryPredictoors
+from util.predictoor.queries import queryPredictoors
```
trentmc, line 23: Why? It's almost always better to use singular not plural.

And this is definitely one of those cases imo

trentmc: I didn't see the other one I guess.

It should be singular everywhere. query.py

util/dftool_module.py
calina-c: I agree with @trizin. That's and advantage of the imports with "from", which allows modularisation. With most frameworks and Python projects it is common to have files with the same name in different modules. If you import those functions mixed in a single entry point, using "from..." is much straightforward than aliasing

util/dftool_module.py
calina-c: Not sure why my previous comment does not appear, it seems to be pending and I can't publish it. Weird.
Here is an example: https://softwareengineering.stackexchange.com/a/75929. Based on this answer, I would argue that a single e.g. "volume" module contains multiple queries, multiple csvs (csv functions), multiple models. Therefore the module is singular because it handles volume-related things, but the "queries" file contains various queries, etc. by analogy models and csvs

util/dftool_module.py
trentmc: OK let's use "from" here and other places where we have >1  module of the same name yet different directory. We agree for sure on that.

For the rest: rather than saying always use one way or another, can we agree to whatever gives the lowest complexity? If a case is about the same for one way vs the other, then I'm fine with either.

util/dftool_module.py
trentmc: Good point, we do have plural in some other places.

So ok then we can call it queries.

Rather than have huge changes in one big PR, can we do this with more incremental PRs? Keep this one as is if you like, but in the future it will be more tractable.

util/predictoor/predictoor_testutil.py
```diff
@@ -0,0 +1,62 @@
+import random
```
calina-c, line 1: should this file reside here, or in predictoor/test?

util/predictoor/predictoor_testutil.py
```diff
@@ -0,0 +1,62 @@
...
+    _weights = [1, 4, 1]
+    _count = 100
+
+    users = [f"0x{i}" for i in range(_count)]
+    stats = {user: {"total": 0, "correct": 0} for user in users}
+
+    def generate_responses(n: int):
+        responses = []
+        for _ in range(n):
+            payouts = [1.0 if random.random() > 0.4 else 0.0 for _ in range(_count)]
```
calina-c, line 18: why not math.floor? floor(random.random) should give the exact same result as the if statement

calina-c (commented):
I'm ready to approve, just tiny suggestions...

util/dftool_module.py
calina-c: I'll handle the change after @trizin merges this PR.

util/predictoor/predictoor_testutil.py
calina-c: well, yes but that's because oceanutil is under util. Predictoor test util should be in the predictoor folder, in my opinion

calina-c: > > Predictoor test util should be in the predictoor folder, in my opinion
> 
> Did you mean it should be in `predicoor/test` folder?

If oceantestutil is in /util, then predictoortestutil should be in predictoor, for consistency. Let's move them later if anything else comes up

calina-c (approved):

util/test/test_dftool_ganache.py
calina-c: I agree with the current approach here. 👍

util/test/predictoor/test_models.py
```diff
@@ -0,0 +1,62 @@
...
+from util.predictoor.models import Prediction
+from util.predictoor.models import Predictoor
```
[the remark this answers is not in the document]
calina-c: this is still pending

calina-c (commented):
There are 3 more unresolved conversations. Should be fairly easy to address

PR #620, opened 2023-05-30
Towards modularising df-py into predictoor/challenge/volume

### LINKED ISSUE #617: Create specific directories for volume DF, predictoor DF, challenge DF
originally referenced here:

              Re the new directories:
- for _this_ issue just move the new predictoor DF-related queries into df-util/predictoor/query.py
- and create a new issue for the rest

_Originally posted by @trentmc in https://github.com/oceanprotocol/df-py/pull/611#discussion_r1209101614_

(NOT: "We should move all raw graphql queries to separate files, for maintainability and readability. They should be grouped under /queries and possibly their purpose, i.e. queries/predictoor_data.py etc. When used, just import them and add the parameters."

calina-c:
Fixes #617.

### DIFF
--- a/util/test/test_dftool_noganache.py
+++ b/util/test/test_dftool_noganache.py
@@ -1,9 +1,10 @@
 import os
 import subprocess
 from typing import List
+from unittest.mock import patch
 
-from enforce_typing import enforce_types
 import pytest
+from enforce_typing import enforce_types
 
 from util import csvs, dftool_module
 
@@ -40,7 +41,7 @@ def test_gen_hist_data():
 @enforce_types
 def test_noarg_commands():
     # Test commands that have no args
-    subargs = _get_HELP_SHORT_subargs_in_dftool()  # key args only, for speed
+    subargs = _get_HELP_subargs_in_dftool("HELP_SHORT")  # key args only, for speed
     subargs = [""] + ["badarg"] + subargs
 
     # these commands are intended to have no parameters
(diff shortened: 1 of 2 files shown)


calina-c:
### moved into volume subfolder (and their tests to volume/test):
	* allocations
	* calcrewards
	* cleancase
	* csvs.py (except for challenge-related csvs which are in /challenge.csvs.py)
	* SimpleDataNft, Tok and TokSet placed in /volume/model.py -> I’m thinking we should pluralise to both models and queries instead of model.py and query.py
	* tousd
	* vesting_schedule
### left in “main”, i.e. under /util itself, under the presumption that these can be used in multiple streams:
	* base18.py (wei conversions)
	* blockrange.py (as far as I can tell, we are only using this in volume but has potential to be used in other places)
	* blocktime.py (same as above)
	* constants.py
	* crypto.py (key, encrypt/decrypt utilities) -> I saw a version of decrypting is used in judge.py (in `_nft_addr_to_pred_vals`) but I did not leverage the function inside crypto.py   in order for that code to stay exactly the same as in the predict-eth repo. I can leverage the function, if needed.
	* csv_helpers.py -> common functionality for csvs
	* dftool_module
	* dispense.py -> I’m not sure about this one. I think we can dispense for other sub streams as well, right?
	* getrate.py -> also unsure about this one, should it be moved to /volume?
	* graphutil.py
	* logger.py
	* multisig.py
	* networkutil.py
	* oceanutil.py
	* oceantestutil.py -> maybe this should reside somewhere in tests?
	* random_addresses.py -> same, probably should be under a test folder somewhere
	* retry.py -> mechanism for retrying functions
### other changes (in challenge substream):
	* I moved some graphql utilities from judge to graphutil.py
	* when splitting csvs.py, I added challenge-related csvs to challenge/csvs.py

PR #637, opened 2023-06-07
Implement Predictoor DF reward calculation

### LINKED ISSUE #635: Implement Predictoor reward calculation
Update calcrewards.py for predictoor_data. Make it easy to toggle on/off, eg by set rewards to 0 by default

Parent epic: [df-py#540](https://github.com/oceanprotocol/df-py/issues/540) "Df-py MVP for Predictoor DF & Challenge DF"

trizin:
Fixes #635 

Changes proposed in this PR:

- Add `calcrewards` module into Predictoor.
- Add predictoor rewards csv functions.
- Add a function to calculate the reward amount given DF week and substream name.
- Unit tests for csv functions.
- Unit tests for `calcrewards` module.
- Unit tests for reward calculation by stream function.
- Add `predictoor` substream into `dftool calc` command.
- Update dispense to check reward CSV file from each substream.
- Unit tests for dftool command.
- Snake-case all csvs functions in the repo.
- Unit tests for dispense with substream csvs and other new functions.

### DIFF
--- a/df_py/util/dftool_module.py
+++ b/df_py/util/dftool_module.py
@@ -373,7 +373,7 @@ def do_predictoor_data():
     CSV_DIR, CHAINID = arguments.CSV_DIR, arguments.CHAINID
 
     # check files, prep dir
-    _exitIfFileExists(predictoor_data_csv_filename(CSV_DIR, CHAINID))
+    _exitIfFileExists(predictoor_data_csv_filename(CSV_DIR))
 
     # brownie setup
     networkutil.connect(CHAINID)


df_py/util/constants.py
```diff
@@ -39,3 +39,4 @@
 # predictoor
 MIN_PREDICTIONS = 1818
 DEPLOYER_ADDRS: List[str] = []
+PREDICTOOR_RELEASE_WEEK = 42  # change me when released
```
calina-c, line 42: I (still) do not agree with adding variables in constants.py, but if it's the only way, I guess it's fine.

df_py/util/dftool_module.py
```diff
@@ -555,70 +568,78 @@ def do_calc():
...
-
-    prev_week = 0
-    if START_DATE is None:
-        cur_week = calcrewards.getDfWeekNumber(datetime.datetime.now())
-        prev_week = cur_week - 1
-    else:
-        prev_week = calcrewards.getDfWeekNumber(START_DATE)
-    m = calcrewards.calcDcvMultiplier(prev_week)
-    print(f"Given prev_week=DF{prev_week}, then DCV_multiplier={m}")
+    if SUBSTREAM_NAME == "volume":
```
calina-c, line 577: I think each of the contents inside these ifs should go into their own modules as functions. i.e. volume/rewards.py, challenge/rewards.py. Dftool_module should only be a thin wrapper to the in-module work, in my opinion.

df_py/predictoor/csvs.py
```diff
@@ -8,6 +11,7 @@
 from df_py.util.csv_helpers import assertIsEthAddr
 
 
+# ------------------------------- PREDICTOOR DATA -------------------------------
 @enforce_types
```
trentmc, line 15: Predictoor is on only one chain, for the foreseeable future. Therefore you don't need to specify chainid in the filename. Instead the filename can simply be "predictoor_data.csv". Similar to Challenge DF.

Currently this cluster of functions for predictoor csv data is around having >1 file. With only one file, it can be simpler. You should (heavily) revise it, patterning after challenge's csvs.py. 

So the functions here are simply:
```python
def save_predictoor_data_csv(predictoor_data, csv_dir: str)
def load_predictoor_data_csv(csv_dir: str) -> Tuple[List[str], List[str], list]:
def predictoor_data_csv_filename(csv_dir: str) -> str:
```

df_py/predictoor/csvs.py
```diff
@@ -70,7 +74,73 @@ def loadPredictoorData(csv_dir: str, chainid: int) -> Dict[str, PredictoorBase]:
...
+
+
 @enforce_types
 def predictoorDataFilename(csv_dir, chainid):
-    f = f"predictoordata_{chainid}.csv"
+    f = f"predictoordata-{chainid}.csv"
+    return os.path.join(csv_dir, f)
+
+
+# ------------------------------- REWARDS -------------------------------
```
trentmc, line 97: Having "OCEAN" in the rewards filename comes from a time when we were refactoring towards rewards from partner teams like H2O / PSDN or Filecoin / FIL. We never completed that because DF volumes never got high enough to justify it. It won't be anytime soon, either. Therefore let's take the opportunity to remove "-OCEAN" from the existing rewards csv filename, and not include "-OCEAN" in the new rewards csv filenames too.

Also, the "rewardsperlp" in "rewardsperlp-OCEAN.csv" assumes that all of DF is Volume DF. We should rename that.

Overall, let's use this PR to set clean csv _rewards_ filenames. I recommend: 
- Volume DF: "volume_rewards.csv" (Rename from rewardsperlp-OCEAN.csv)
- Predictoor DF: "predictoor_rewards.csv"
- Challenge DF: "challenge_rewards.csv"

Let's also have clean csv _data_ filenames. I recommend:
- Volume DF: has many files. Leave them as-is.
- Predictoor DF: call it "predictoor_data.csv" (I recommend this in a comment above as well)
- Challenge DF: call it "challenge_data.csv"

df_py/predictoor/csvs.py
```diff
@@ -70,7 +74,73 @@ def loadPredictoorData(csv_dir: str, chainid: int) -> Dict[str, PredictoorBase]:
...
+def loadPredictoorRewards(
+    csv_dir: str, token_symbol: str = "OCEAN"
+) -> Dict[str, float]:
+    csv_file = predictoorRewardsFilename(csv_dir, token_symbol)
+    predictoor_rewards = {}
+    with open(csv_file, "r") as f:
+        reader = csv.reader(f)
+        for row_i, row in enumerate(reader):
+            if row_i == 0:
+                assert row == ["predictoor_addr", "reward"]
```
trentmc, line 130: Change `"reward"` to `"OCEAN_amt"` for consistency with volume DF rewards csv.

df_py/util/dftool_module.py
```diff
@@ -54,7 +65,7 @@
   dftool vebals ST FIN NSAMP CSV_DIR CHAINID [RETRIES]
   dftool challenge_data CSV_DIR [DEADLINE] [RETRIES]
   dftool predictoor_data CSV_DIR START_DATE END_DATE CHAINID [RETRIES]
-  dftool calc CSV_DIR TOT_OCEAN [START_DATE] [IGNORED] - from stakes/etc csvs, output rewards csvs across Volume + Challenge + Predictoor DF
+  dftool calc CSV_DIR TOT_OCEAN START_DATE [SUBSTREAM_NAME] [IGNORED] - from stakes/etc csvs, output rewards csvs across Volume + Challenge + Predictoor DF
```
trentmc, line 68: I really like you you put aggregation of csv rewards across substreams into `dftool dispense_active`. Makes lots of sense, and the implementation looks good.

----

Now that I see this implementation, I realize it's much cleaner to have a _different_ dftool command for each substream. 

I propose:
```
dftool calc_volume_rewards ...
dftool calc_challenge_rewards ...
dftool calc_predictoor_rewards ...
```

(I see that Calina had a similar comment. Good! :)

----

Small thing: you can remove the "IGNORED" stuff. That's only there because I didn't want to cause downstream breakage in df-sql. But you're making enough changes that you have to change/test downstream anyway. Perhaps as you make the changes above, then "IGNORED" will naturally get tossed out.

df_py/util/dftool_module.py
trentmc: Agree

df_py/util/dftool_module.py
```diff
@@ -681,14 +700,27 @@ def do_dispense_active():
     from_account = _getPrivateAccount()
     token_symbol = B.Simpletoken.at(TOKEN_ADDR).symbol().upper()
     token_symbol = token_symbol.replace("MOCEAN", "OCEAN")
-    rewards = csvs.loadRewardsCsv(CSV_DIR, token_symbol)
 
-    # "flatten" the rewards dict to dispense all chains in one go
-    all_rewards = calcrewards.flattenRewards(rewards)
+    rewards = {}
```
trentmc, line 704: It's weird to start with an empty `rewards` dict then update it with subsequent calls to calc rewards for each substream.

Rather, each substream should calc its own rewards from its budget. 

It'd look something like:
```python
volume_rewards = load_volume_rewards_csv(...)
predictoor_rewards = load_predictoor_rewards_csv(...)
challenge_rewards = load_challenge_rewards_csv(...)
rewards = merge_rewards(volume_rewards, predictoor_rewards, challenge_rewards)
```

----

In an earlier comment I mention how we don't need to specify "OCEAN" token in rewards filenames anymore. Therefore we don't need "OCEAN" arguments in the function calls below.

df_py/predictoor/csvs.py
```diff
@@ -1,5 +1,8 @@
 import csv
 import os
+import glob
+import re
+
```
trentmc, line 5: (Commenting here because it's hard to insert comments into files that currently have no changes)

In df_py/volume/csvs.py, we should rename the functions to (a) make the "volume" part explicit just as "predictoor" and "challenge" is explicit, and (b) ensure snake case. That is:
```
saveRewardsperlpCsv --> save_volume_rewards_csv
loadRewardsCsv --> load_volume_rewards_csv
rewardsperlpCsvFilename --> volume_rewards_csv_filename

saveRewardsinfoCsv --> save_volume_rewardsinfo_csv
rewardsinfoCsvFilename --> volume_rewardsinfo_csv_filename
```

----

While you're at it, make the other functions in df_py/volume/csvs.py snake case too.

----

While you're at it, make the functions in df_py/challenge/csvs.py snake case too.

trentmc (changes requested):

df_py/predictoor/test/test_predictoor_csvs.py
```diff
@@ -47,3 +47,35 @@ def test_predictoordata(tmp_path):
             loaded_predictoor.correct_prediction_count
             == original_predictoor.correct_prediction_count
         )
+
+
+@enforce_types
+def test_predictoorrewards(tmp_path):
```
calina-c, line 53: I suggest `test_predictoor_rewards`

df_py/predictoor/test/test_predictoor_csvs.py
```diff
@@ -47,3 +47,35 @@ def test_predictoordata(tmp_path):
...
+    for i in range(5):
+        predictoor_rewards[f"0x{i}000000000000000000000000000000000000000"] = (
+            i + 1
+        ) * 10.0
+
+    csv_dir = str(tmp_path)
+    csvs.save_predictoor_rewards_csv(predictoor_rewards, csv_dir)
+
+    with open(csvs.predictoor_rewards_csv_filename(csv_dir), "r") as loaded_data:
+        data = loaded_data.read().strip()
```
calina-c, line 72: here is an instance where csv dictreader and dictwriter would have helped. The test data could simply be a dictionary, not a string and loaded like this

df_py/util/constants.py
calina-c, line 42: can we say `# change me when Predictoor is released`? The wording makes me believe weird stuff I've described on Slack 😄

df_py/util/dftool_module.py
```diff
@@ -166,9 +179,9 @@ def do_volsym():
     (Vi, Ci, SYMi) = retryFunction(
         queries.queryVolsOwnersSymbols, RETRIES, 60, rng, CHAINID
     )
-    csvs.saveNftvolsCsv(Vi, CSV_DIR, CHAINID)
-    csvs.saveOwnersCsv(Ci, CSV_DIR, CHAINID)
-    csvs.saveSymbolsCsv(SYMi, CSV_DIR, CHAINID)
+    csvs.save_nftvols_csv(Vi, CSV_DIR, CHAINID)
```
calina-c, line 182: these name changes will create conflicts with the args branch, but we'll sort it out. I think they are beneficial, so let's keep them.

calina-c (changes requested):

df_py/util/dftool_module.py
```diff
@@ -509,30 +522,46 @@ def do_predictoor_data():
...
 @enforce_types
-def do_calc():
-    HELP = """from stakes/etc csvs, output rewards csvs across Volume + Challenge + Predictoor DF
+def do_calc_predictoor_rewards():
+    _do_calc("predictoor")
 
-Usage: dftool calc CSV_DIR TOT_OCEAN [START_DATE] [IGNORED]
+
+@enforce_types
+def _do_calc(substream_name):
```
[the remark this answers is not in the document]
calina-c: Hmmm... I would have SUBSTREAM_NAME as a parameter, then. What do you think? `dftool calc_rewards SUBSTREAM_NAME CSV_DIR ...`

df_py/util/dftool_module.py
calina-c: I did have a similar comment, and the underlying functionality implemented by Berkay looks great!

But in regards to having different commands, I think would better with a parameter. i.e. `dftool calc_rewards volume...` not, `dftool calc_volume_rewards...` because it makes the handling really ugly and if we can have a parameter instead of a new function it seems to make more sense.

df_py/util/dftool_module.py
calina-c: Closing this one as well, since there is another comment on this topic.

df_py/util/dftool_module.py
calina-c: Fixed. I am marking this as resolved, hope I'm not overstepping.

calina-c (approved):

calina-c (commented):
could we remove chain id from the arguments as well?

calina-c (approved):

calina-c:
> > could we remove chain id from the arguments as well?
> 
> Where's that?

Sorry, I thought it got attached to the correct line. I thought about making CHAINID an optional argument for the command itself, but I changed my mind. We should keep it for consistency and future implementations.

trentmc (approved):

PR #639, opened 2023-06-08
Adds argument parsing

calina-c:
Work on #640.

Accomplishments:
- tighter argument parsing, no need to count arguments manually
- optional arguments are now actually options, not arguments -> easier management of defaults and no more reliance on ordering and other missing/present args/options
- automatic validation of parameters through reusable functions e.g. for ST and FIN blocks, for CSV paths (and auto-created if needed)
- reusable classes of arguments for commands that share the same set of arguments and options
- automatic help formatting: no need for HELP strings since argparse shows the exact needed thing if an argument is missed
one-liner parameter printing, e.g. `dftool volsym -h`, `dftool volsym --help`
- argument parsing is in a separate file, improving readability and Separation of Concerns principle for dftool_module
- fixed some older typos and unused variables (which we did not realise at the time because the code was more complex)

BONUS/metrics: dftool_module is ~330 lines shorter, with BETTER functionality than before.

I have NOT touched `calc` to avoid conflicts. We need to convert that one too. That is why I marked my PR as "work on" instead of "closes".

### DIFF
--- a/.github/workflows/flow.yml
+++ b/.github/workflows/flow.yml
@@ -119,14 +119,14 @@ jobs:
           USE_MULTISIG: true
         if: github.event_name != 'schedule'
         run: |
-          ./dftool dispense_active $CSV_DIR 5 0x6fd867E5AEE6D62a24f97939db90C4e67A73A651 0xCfDdA22C9837aE76E0faA845354f33C62E03653a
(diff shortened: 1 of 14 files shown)


calina-c:
> Could you update the scripts in `df-systems` folder? These scripts are set to run on the cloud instance as per the schedule.

I've checked all and there is no need to change them because they did not define optional arguments, i.e. retries, etc.
My only concern is that in helperscript we seem to have a command called calculate_passive and I can not find that command.

calina-c:
> Awesome work! Although the total coverage has been increased, is it fine to have 17.7% coverage in this PR? I guess it's related to new argument parser classes.

Yeah, I'll work on covering more parsing edge cases as part of increasing test coverage, to avoid doubling the work.

PR #651, opened 2023-06-16
Use latest release of df-py in reward dispense flows

### LINKED ISSUE #628: Update reward distribution CI to use latest release
Update reward distribution CI to use the latest release of df-py (vs main branch which is status quo)

trizin:
Fixes #628

Changes proposed in this PR:

- Use latest release of df-py in reward dispense flows.

### DIFF
--- a/df_py/challenge/csvs.py
+++ b/df_py/challenge/csvs.py
@@ -83,7 +83,7 @@ def load_challenge_data_csv(csv_dir: str) -> Tuple[List[str], List[str], list]:
 
 @enforce_types
 def challenge_data_csv_filename(csv_dir: str) -> str:
-    f = "challenge.csv"
+    f = "challenge_data.csv"
     return os.path.join(csv_dir, f)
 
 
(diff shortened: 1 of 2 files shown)


.github/workflows/checkpoint_feedist.yml
```diff
@@ -14,7 +14,20 @@ jobs:
...
+      - id: df-py
+        uses: pozetroninc/github-action-get-latest-release@master
+        with:
+          owner: oceanprotocol
+          repo: df-py
+          excludes: prerelease, draft
+
+      - name: Set branch to checkout
+        id: set_branch
+        run: echo "BRANCH=${{ github.event_name == 'schedule' && steps.df-py.outputs.release || github.ref }}" >> $GITHUB_ENV
```
calina-c, line 26: I think it's possible to use this directly in checkout, without setting it separately. You will have ref: ${{ steps.df-py.outputs.release }}

calina-c: Ok then here is what I suggest:
- set a default branch to handle the null setting. I don't know how it behaves if you do not set the env, but specify the ref key
- also, I believe BRANCH is a bit generic, can you name it to something more specific? E.g. DF_PY_LATEST_BRANCH or something similar?

calina-c (changes requested):

calina-c (approved):

Issue #618, opened 2023-05-29
Implement Pythonic casing standards


calina-c:
Currently there is a mixture of snake_case and camelCase in the repo. As it grows, we should implement standard Python casing, which is CamelCase for classes and snake_case for functions.

PR #618 is an input to this. (Don't merge it)

PR #506, opened 2023-03-15
Fix dftool commands to be snake_case and consistent with codebase

idiom-bytes:
Fixes #505 

### DoD:
- [ ] Dev Ops - DO NOT MERGE => Update/deploy all workflows. Consumers of df-py could break. i.e. shell scripts around dfsql that call dftool commands.

### Changes proposed in this PR:

**Rename functions**
- Rename dftool commands to be snake_case and consistent: init_dev_wallets
- Except when using system objects and camelCase functions, then keep it consistent: `new_veOcean`
- Except when using existing two-syllable commands, then keep it consistent: `getrate` and `vebals`

**Ignore artifacts**
- Disable .coverage and .vecode/ artifact folders in gitignore
- Moved to https://github.com/oceanprotocol/team-sailfish/issues/41

PR #655, opened 2023-06-26
Fixes Pythonic casing in commands and function names

calina-c:
Work on #618. Supersedes #506, which can be closed after merging this one.

Still pending:
- [ ] questions about nftinfo (should it be nft_info? should we also change the csv file name?)
- [ ] capitalisation in test files (only modules are covered up until now)


calina-c:
I updated most commands to snake case and superseding #506. Much easier now that we have tests.
I think the only two pending are nftinfo and getrate. Should I convert those too? And, if I do convert nftinfo to nft_info, should I also change the csv filenames?

PR #660, opened 2023-06-28
Use OCEAN for challenge DF rewards

### LINKED ISSUE #658: Use OCEAN in Challenge DF reward calculations
5000 OCEAN from active rewards budget are dedicated to Challenge DF each week

trizin:
Fixes #658

Changes proposed in this PR:

- Use OCEAN for challenge DF rewards


df_py/challenge/calc_rewards.py
```diff
@@ -15,12 +15,14 @@ def get_challenge_reward_amounts_in_usdt(
     @return
       list of USDT amounts, in order of 1st, 2nd, 3rd place
     """
+    rewards_in_ocean = get_challenge_reward_amounts_in_ocean(at_date)
```
trentmc, line 18: Why do you even need the method "get_challenge_reward_amounts_in_usdt()"?

df_py/challenge/csvs.py
```diff
@@ -83,7 +83,7 @@ def load_challenge_data_csv(csv_dir: str) -> Tuple[List[str], List[str], list]:
 
 @enforce_types
 def challenge_data_csv_filename(csv_dir: str) -> str:
-    f = "challenge.csv"
+    f = "challenge_data.csv"
```
trentmc, line 86: I'm surprised this is being changed here. It's outside the scope of this issue.

And if this change isn't breaking things outside of df-py (in df-sql), maybe it's a sign that we need to test more throughly.

trentmc (changes requested):

calina-c (approved):
I thought we should send rewards in USDT but this is MUCH simpler if all right with everyone.

trentmc (approved):

Issue #629, opened 2023-06-05
Update modules to use block time instead of local time


trizin:
Remove usages of local/machine time and replace them with chain time.

calina-c:
Adding more context here and expanding the scope a bit. Currently we can not run tests in any given order, meaning they are interdependent somehow. In order to advance things, we added a string ordering of test modules in the test GitHub Action. However, we commented out the blocktime tests, which only work if ran before any advances to the chain.

I think this issue should include:
- removing dependencies between tests, allow for any ordering, i.e. simply run `pytest util` instead of `pytest util/volume util/challenge [...]`. -> this is brittle because if new modules are added someone might forget to add them to the flow.
- unskipping and fixing blocktime tests, where usages of chain time should be employed.

PR #666, opened 2023-06-28
Fix randomly failing passive reward query test

### LINKED ISSUE #665: Randomly failing passive rewards test
https://github.com/oceanprotocol/df-py/actions/runs/5401832201/jobs/9812210444

This test is randomly failing

trizin:
Fixes #665

Changes proposed in this PR:

- Fix randomly failing passive reward query test

### DIFF
--- a/df_py/challenge/calc_rewards.py
+++ b/df_py/challenge/calc_rewards.py
@@ -4,23 +4,6 @@ from typing import Any, Dict, List, Optional
 from enforce_typing import enforce_types
 
 from df_py.util.constants import CHALLENGE_FIRST_DATE
-from df_py.util.get_rate import get_rate
-
-
-@enforce_types
-def get_challenge_reward_amounts_in_usdt(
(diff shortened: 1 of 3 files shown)


df_py/volume/test/test_queries.py
```diff
@@ -484,7 +484,7 @@ def sim_epoch():
             to_wei(1000.0),
             {"from": god_acct},
         )
-        chain.sleep(WEEK)
+        chain.sleep(WEEK * 2)
```
calina-c, line 487: could this be related to #629?

calina-c:
Here's an interesting thing: I ran some experiments to remove the custom ordering of tests (that we define in test.yml). It seems to work except for this same test (and another blocktime one). See this run here: https://github.com/oceanprotocol/df-py/actions/runs/5409841094/jobs/9831088475

I think we could remove the custom ordering after we figure out what is happening here.

calina-c (approved):

PR #680, opened 2023-07-11
Update predictoor reward distribution

### LINKED ISSUE #679: Update predictoor reward distribution
Remove accuracy reward distribution and update DCV boundary for predictoor assets

trizin:
Fixes #679

Changes proposed in this PR:

- 20% multiplier for predictoor assets.
- Added a new parameter `contract_multipliers` into Volume DF calc function.
- Tests for the new parameter and 20% predictoor multiplier.
- Removed old predictoor reward distribution code.

### DIFF
--- a/df_py/util/dftool_module.py
+++ b/df_py/util/dftool_module.py
@@ -317,6 +317,13 @@ def do_challenge_data():
             Example for Round 5: 2023-05-03_23:59
         """,
     )
+    parser.add_argument(
+        "--RETRIES",
+        default=1,
+        type=int,
+        help="# times to retry failed queries",
+        required=False,
+    )
 
     arguments = parser.parse_args()
     print_arguments(arguments)
@@ -335,7 +342,9 @@ def do_challenge_data():
 
     # main work
     deadline_dt = judge.parse_deadline_str(arguments.DEADLINE)
-    challenge_data = judge.get_challenge_data(deadline_dt, judge_acct)
+    challenge_data = retry_function(
+        judge.get_challenge_data, arguments.RETRIES, 10, deadline_dt, judge_acct
+    )
 
     save_challenge_data_csv(challenge_data, csv_dir)
 
--- a/.github/workflows/flow.yml
+++ b/.github/workflows/flow.yml
@@ -2,7 +2,7 @@ name: Dispense Active Rewards
 
 on:
   schedule:
-    - cron: "0 0 * * 4"
+    - cron: "2 1 * * 4"
   push:
     branches: [main]
   pull_request:
@@ -115,6 +115,16 @@ jobs:
         run: |
           cp $CSV_DIR/rate-OCEAN.csv $CSV_DIR/rate-MOCEAN.csv
           sed -i -e 's/MOCEAN/OCEAN/g' $CSV_DIR/rate-MOCEAN.csv
+
+      - name: Challenge DF data
+        run: |
+          ./dftool challenge_data $CSV_DIR --RETRIES 3
+        env:
+          JUDGE_PRIVATE_KEY: ${{secrets.JUDGE_PRIVATE_KEY}}
+        
+      - name: Run dftool calc rewards for DF Challenge
+        run: |
+          ./dftool calc challenge $CSV_DIR 0 --START_DATE $date
 
       - name: Run dftool calc rewards for volume
         run: |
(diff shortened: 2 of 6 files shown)


df_py/volume/test/test_calcrewards.py
```diff
@@ -963,6 +993,61 @@ def test_calc_rewards_volume():
...
+    ), patch(
+        "os.path.exists",
+        return_value=True,
+    ), patch(
+        "df_py.volume.calc_rewards.load_predictoor_contracts_csv",
+        return_value=mock_data["predictoor_contracts"],
+    ), patch(
+        "df_py.volume.calc_rewards.get_df_week_number", return_value=30
+    ):
+        rewards_per_lp, rewards_info = calc_rewards_volume(
```
idiom-bytes, line 1040: Had not seen patch() before, nice implementation.

df_py/volume/calc_rewards.py
```diff
@@ -467,6 +482,12 @@ def calc_rewards_volume(
     SYM = csvs.load_symbols_csvs(CSV_DIR)
     R = csvs.load_rate_csvs(CSV_DIR)
 
+    contract_multipliers: Dict[str, float] = {}
+
+    if os.path.exists(predictoor_contracts_csv_filename(CSV_DIR)):
+        print("Found predictoor contracts")
+        predict_contracts = load_predictoor_contracts_csv(CSV_DIR)
+        contract_multipliers = {i: 0.2 for i in predict_contracts.keys()}
```
idiom-bytes, line 490: Perhaps change 0.2 => PREDICTOOR_MULTIPLIER?

df_py/util/dftool_module.py
```diff
@@ -389,6 +385,12 @@ def do_predictoor_data():
     st_block, fin_block = get_st_fin_blocks(chain, arguments.ST, arguments.FIN)
 
     # main work
+    predictoor_contracts = query_predictoor_contracts(chain_id)
+
+    # retry_function(
+    #     query_predictoor_contracts, arguments.RETRIES, 10, chain_id
+    # )
```
idiom-bytes, line 392: Not sure whether to comment on this or not. I'm sure this is for debugging purposes... maybe wrap it in an arg?

i.e. `do_predictoor_data(retry = false)`

idiom-bytes (approved):
Left some comments, looks good. Didn't go very deep into calcs but tests provided a good overview.

df_py/util/vesting_schedule.py
```diff
@@ -26,8 +26,7 @@ def get_active_reward_amount_for_week_eth_by_stream(
     dfweek = get_df_week_number(start_dt) - 1
 
     if substream == "predictoor":
-        # 0.01%
-        return total_reward_amount * 0.001 if dfweek >= PREDICTOOR_RELEASE_WEEK else 0
+        return 35000 if dfweek >= PREDICTOOR_RELEASE_WEEK else 0
```
calina-c, line 29: I think this 35000 should be more visible, in the form of a constant or something similar. Also, can you please format it as `35_000` for better readability?

df_py/volume/calc_rewards.py
```diff
@@ -143,16 +154,18 @@ def _stake_vol_dicts_to_arrays(
...
     for j, (chainID, nft_addr) in enumerate(chain_nft_tups):
         for i, LP_addr in enumerate(LP_addrs):
             assert nft_addr in stakes[chainID], "each tup should be in stakes"
             S[i, j] = stakes[chainID][nft_addr].get(LP_addr, 0.0)
         V_USD[j] += nftvols_USD[chainID].get(nft_addr, 0.0)
+        if nft_addr in contract_multipliers:
+            M[j] = contract_multipliers[nft_addr]
 
-    return S, V_USD
+    return S, V_USD, M
```
calina-c, line 168: I am not a fan of naming variables with such names. It does seem the entire flow is like this, and we can leave it as such but we should think about capitalisation and length (per suggestive names) in these cases, especially since I see it gets injected with the same name as a parameter for other functions.

calina-c (changes requested):

df_py/volume/calc_rewards.py
trentmc: Please keep these names as-is. These names are short because they're repeated so many places, and longer names makes the code _way_ harder to read. (I discovered this trick doing hardcore numerical processing. It applies elsewhere too.)

It's all about managing complexity:)

df_py/util/vesting_schedule.py
trentmc: Fully agree. We should avoid having magic numbers randomly scattered throughout the code. Rather, put them in https://github.com/oceanprotocol/df-py/blob/main/df_py/util/constants.py. 

Same  goes for other magic numbers.

calina-c (approved):

PR #734, opened 2023-11-03
#731 - Set dcv multiplier to 0.201 for OPF deployed Predictoor contracts

### LINKED ISSUE #731: Set DCV multiplier to 0.201 for OPF deployed predictoor contracts
Set DCV multiplier to 0.201 for OPF deployed predictoor contracts

Related issue: [pdr-private#67](https://github.com/oceanprotocol/df-private/issues/67) to ensure correct mental model & calculations wrt DCV.

trizin:
Fixes #731

Changes proposed in this PR:

- Set DCV multiplier to 0.201 for OPF predictoor assets.
- Update tests accordingly.

### DIFF
--- a/.bumpversion.cfg
+++ b/.bumpversion.cfg
@@ -1,6 +1,0 @@
-[bumpversion]
-current_version = 0.0.4
-commit = True
-tag = True
-
-[bumpversion:file:setup.py]
--- a/Dockerfile
+++ b/Dockerfile
@@ -7,6 +7,7 @@ WORKDIR /app/df-py
 RUN python3.8 -m pip install --upgrade pip
 RUN python3.8 -m pip install .
 
+RUN apt-get update && apt-get install -y npm
 RUN npm install @openzeppelin/contracts
 ENV MUMBAI_RPC_URL="https://polygon-mumbai.infura.io/v3/"
 ENV OASIS_SAPPHIRE_RPC_URL="https://sapphire.oasis.io"
--- a/README.md
+++ b/README.md
@@ -256,3 +256,7 @@ How to fix:
 - Tactic: in barge console: `./cleanup.sh`
 - Tactic: in any console: `rm -rf ~/.ocean`
 - More tactics at [barge README](https://github.com/oceanprotocol/barge)
+
+
+## Release process (developers)
+In order to release a new version of df-py, simply create a new GitHub release. As opposed to other Ocean Protocol python repos, there is no need to use bumpversion.
--- a/setup.py
+++ b/setup.py
@@ -30,7 +30,6 @@ install_requirements = [
     "requests",
     "scipy",
     "types-requests",
-    "bumpversion",
     "black",
     "ocean-contracts==2.0.0a15",
 ]


df_py/util/constants.py
```diff
@@ -41,7 +41,7 @@
     23295: ["0xe02a421dfc549336d47efee85699bd0a3da7d6ff"],
 }
 PREDICTOOR_RELEASE_WEEK = 63
-PREDICTOOR_MULTIPLIER = 0.2
+PREDICTOOR_MULTIPLIER = 0.201
```
trentmc, line 44: Two general comments.

First: in `volume/calc_rewards.py` there's one function called `calc_rewards()_volume()` which then calls 'calc_rewards()` and directly returns its results. That's super confusing, and gets in the way of the next issues / fixes. Please rename these to be more precise.

Here's an idea: simply rename `calc_rewards()` to `_calc_rewards_volume()`. The `volume` makes it clear that it's about volume (vs general), the `_` makes it clear that it is not meant for external consumption (vs `calc_rewards_volume()` which is).

Second: The code around dcv_multiplier in `calc_rewards.py` & `constants.py` is too messy. It needs cleaning up. Details of issue:
- theres `calc_dcv_multiplier()` which was used as a soft ramp from inf to 1.0 to a value of 0.001. (And its final return value is 0.03 but is never used. Another bug.) It returns a scalar.
  - that scalar is passed into `calc_rewards()` then `_calc_rewards_USD()`
- then there's `contract_multipliers` dict which is is passed into `calc_rewards()`
  - then it's converted into an array M
  - which is then passed to  `_calc_rewards_USD()`
- and then there's more logic for multipliers in `calc_rewards_volume()`. 
  - whereas most of `calc_rewards_volume()` simply loads 5 CSVs of pre-calculated values, taking one line each, predictoor stuff takes up 7 lines (way too much real estate).
  - and ideally, it shouldn't have to detect whether predictoor csvs exist. 
- In short, with the logic scattered everywhere, it's hard to know what's happening

Re this messiness, I'd like to see all dcv_multiplier stuff in one place, or max two.

I recommend:
- `calc_rewards()` should input `DFweek`, not `DCV_multiplier`. 
- `_calc_rewards_USD()` should input `DFweek`, not `DCV_multiplier`. 
- inside `_calc_rewards_USD()`, within the loop for a given asset, it calls `calc_DCV_multiplier(DFweek:int, isPredictoor: bool)`
- then change `calc_DCV_multiplier()` logic to support the additional argument. Just hardcode the 0.201 there. Don't bother with PREDICTOOR_MULTIPLIER. 

Also:
-  `_calc_rewards_USD()` and its caller (`calc_rewards()`), and its caller (`calc_rewards_volume()`), will also need `isPredictoor` values. Basically, replace `M` (1d array) with `P` (1d array) (= is this nft a predictoor), and `contract_multiplier` (dict) with `predictoor_addrs' (set of addr).

I recognize that this will take a bit more work, and tests to be updated. But the result will be much cleaner and easier to understand

trentmc (changes requested):

df_py/util/constants.py
trentmc: OK

PR #743, opened 2023-11-09
Calculate and include timeframe and source in predictoor asset name

### LINKED ISSUE #742: Add timeframe to Predictoor Assets name
Right now it's hard to differentiate on the df-web side which is a 1h timeframe and which is a 1d timeframe asset. They are looking exactly the same.
I propose we also add the timeframe to the asset name so we know what's the difference.

trizin:
Fixes #742

Changes proposed in this PR:

- Calculate and include timeframe and source in predictoor asset name
- 
-

### DIFF
--- a/df_py/predictoor/queries.py
+++ b/df_py/predictoor/queries.py
@@ -125,7 +125,7 @@ def query_predictoor_contracts(chain_id: int) -> Dict[str, PredictContract]:
             timeframe = info["timeframe"]
             source = info["source"]
 
-            asset_name = f"{pair}-{timeframe}-{source}"
+            asset_name = f"{pair}-{source}-{timeframe}"
 
             contract_obj = PredictContract(
                 chain_id,


df_py/predictoor/queries.py
```diff
@@ -70,10 +119,18 @@ def query_predictoor_contracts(chain_id: int) -> Dict[str, PredictContract]:
                 continue
 
             nft_addr = contract["token"]["nft"]["id"]
+            info725 = contract["token"]["nft"]["nftData"]
+            info = info_from_725(info725)
+            pair = info["pair"]
+            timeframe = info["timeframe"]
+            source = info["source"]
+
+            asset_name = f"{pair}-{timeframe}-{source}"
```
trentmc, line 128: Please make it `f"{pair}-{source}-{timeframe}"`

(It should be like this everywhere, btw. Since pair and source are intrinsically bound to each other.)

trentmc (changes requested):

trentmc (approved):
lgtm

PR #769, opened 2023-11-21
#724 - Predictoor automate dispensing predictoor df

### LINKED ISSUE #724: [predictoor] Dispense predictoor DF
Right now, every Friday, we do "dispense passive rewards" (75k)

We also need to dispense predictoor DF rewards (37k) towards 0x086E7F0588755af5AF5f8194542Fd8328238F3C1  (on Ethereum)

During the week, those Ocean tokens will be bridged to Sapphire and then sent to DFBuyer.

trizin:
Fixes #724

Changes proposed in this PR:

- New dftool command `do_dispense_predictoor` + test
- Multisig function to transfer erc tokens to any address + test
- Renamed dispense passive to dispense rewards and added dispense predictoor command to dispense rewards every Friday.

### DIFF
--- a/.github/workflows/dispense_predictoor_rose.yml
+++ b/.github/workflows/dispense_predictoor_rose.yml
@@ -103,3 +103,26 @@ jobs:
         if: github.event_name == 'schedule'
         run: |
           ./dftool dispense_active $CSV_DIR 23294 --DFREWARDS_ADDR="0xc37F8341Ac6e4a94538302bCd4d49Cf0852D30C0" --TOKEN_ADDR="0x8Bc2B030b299964eEfb5e1e0b36991352E56D2D3" --PREDICTOOR_ROSE True
+
+      - name: Rename folder to round number
+        run: |
+          bash ./scripts/workflow_rename_folder.sh $now
+
+      - id: "auth"
+        uses: "google-github-actions/auth@v1"
+        with:
+          credentials_json: "${{ secrets.GC_SECRET }}"
+
+      - name: Set upload destination
+        run: |
+          if [[ "${{ github.event_name }}" == "schedule" ]]; then
+            echo "DESTINATION=df-historical-data" >> $GITHUB_ENV
+          else
+            echo "DESTINATION=df-historical-data-test" >> $GITHUB_ENV
+          fi
+
+      - id: "upload-folder"
+        uses: "google-github-actions/upload-cloud-storage@v1"
+        with:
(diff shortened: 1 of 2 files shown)


.github/workflows/dispense_rewards.yml
```diff
@@ -73,3 +73,17 @@ jobs:
         if: github.event_name == 'schedule'
         run: |
           ./dftool dispense_passive 1 0 $date
+
```
trentmc, line 76: The filenames in this directory are confusing. Renaming this file from dispense_passive -> dispense_active makes it worse. We need to get things consistent.

Currently, here are the relevant files, and their function:
- dispense_passive.yml / dispense_rewards.yml : Passive DF & Predictoor DF OCEAN?
- flow.yml: says "Dispense Active Rewards" but I think it's for Volume DF & Challenge DF?
- predictoor_rose.yml: rewards for Predictoor DF ROSE? 

Perhaps what's better:
- dispense_passive.yml : Passive DF
- dispense_volume.yml: Volume DF & Challenge DF (we could have had "challenge" in the name. But we'll turn off "challenge" shortly anyway; therefore this minimizes changes)
- dispense_predictoor_ocean.yml : Predictoor DF OCEAN
- dispense_predictoor_rose.yml: Predictoor DF ROSE

df_py/util/dftool_module.py
```diff
@@ -1084,6 +1084,41 @@ def do_dispense_passive():
     print("Dispensed passive rewards")
 
 
+# ========================================================================
+
+
+@enforce_types
+def do_dispense_predictoor():
```
trentmc, line 1091: Similar to the github actions filenames, we need better clarity on these method names.

It should be something like:
- do_dispense_passive
- do_dispense_volume (includes challenge, for now)
- do_dispense_predictoor_ocean
- do_dispense_predictoor_rose

OR
- do_dispense_passive
- do_dispense_active, which dispenses volume, challenge, predictoor ocean, and predictoor rose

trentmc (changes requested):

df_py/util/dftool_module.py
trentmc: Thanks for the info. 

OK we don't need `do_dispense_predictoor_rose`. Not worth it.

It seems that `do_dispense_predictoor_ocean` needs a better label, since it "Funds the account which will to bridge OCEAN to Sapphire" versus actually dispensing rewards.

trentmc (approved):

PR #770, opened 2023-11-22
Apply OOP to volume reward calculations.

### LINKED ISSUE #741: Volume reward calculation functions are cryptic
Context: https://github.com/oceanprotocol/df-py/pull/734#discussion_r1381826296

Trent:
First: in `volume/calc_rewards.py` there's one function called `calc_rewards()_volume()` which then calls 'calc_rewards()` and directly returns its results. That's super confusing, and gets in the way of the next issues / fixes. Please rename these to be more precise.

Here's an idea: simply rename `calc_rewards()` to `_calc_rewards_volume()`. The `volume` makes it clear that it's about volume (vs general), the `_` makes it clear that it is not meant for external consumption (vs `calc_rewards_volume()` which is).

Second: The code around dcv_multiplier in `calc_rewards.py` & `constants.py` is too messy. It needs cleaning up. Details of issue:
- theres `calc_dcv_multiplier()` which was used as a soft ramp from inf to 1.0 to a value of 0.001. (And its final return value is 0.03 but is never used. Another bug.) It returns a scalar.
  - that scalar is passed into `calc_rewards()` then `_calc_rewards_USD()`
- then there's `contract_multipliers` dict which is is passed into `calc_rewards()`
  - then it's converted into an array M
  - which is then passed to  `_calc_rewards_USD()`
- and then there's more logic for multipliers in `calc_rewards_volume()`. 
  - whereas most of `calc_rewards_volume()` simply loads 5 CSVs of pre-calculated values, taking one line each, predictoor stuff takes up 7 lines (way too much real estate).
  - and ideally, it shouldn't have to detect whether predictoor csvs exist. 
- In short, with the logic scattered everywhere, it's hard to know what's happening

Re this messiness, I'd like to see all dcv_multiplier stuff in one place, or max two.

I recommend:
- `calc_rewards()` should input `DFweek`, not `DCV_multiplier`. 
- `_calc_rewards_USD()` should input `DFweek`, not `DCV_multiplier`. 
- inside `_calc_rewards_USD()`, within the loop for a given asset, it calls `calc_DCV_multiplier(DFweek:int, isPredictoor: bool)`
- then change `calc_DCV_multiplier()` logic to support the additional argument. Just hardcode the 0.201 there. Don't bother with PREDICTOOR_MULTIPLIER. 

Also:
-  `_calc_rewards_USD()` and its caller (`calc_rewards()`), and its caller (`calc_rewards_volume()`), will also need `isPredictoor` values. Basically, replace `M` (1d array) with `P` (1d array) (= is this nft a predictoor), and `contract_multiplier` (dict) with `predictoor_addrs' (set of addr).

calina-c:
Closes #741.


df_py/volume/reward_calculator.py
```diff
@@ -0,0 +1,479 @@
...
+                -1
+                if owner_addr not in self.LP_addrs
+                else self.LP_addrs.index(owner_addr)
+            )
+
+        return S, V_USD, M, C
+
+    @freeze_attributes
+    @enforce_types
+    def _calc_usd(self) -> np.ndarray:
```
trentmc, line 152: Rename to _calc_rewards_usd().

(It should have always been this, but now's a great time to do so.)

df_py/volume/reward_calculator.py
```diff
@@ -0,0 +1,479 @@
...
+
+
+def _calc_dcv_multiplier(DF_week: int) -> float:
+    if DF_week < 9:
+        return np.inf
+
+    if 9 <= DF_week <= 28:
+        return -0.0485 * (DF_week - 9) + 1.0
+
+    return 0.001
```
trentmc, line 444: It's confusing to have two routines, with names "calc_dcv_multiplier" and "_calc_dcv_multiplier".

I recommend to merge into a single routine.

df_py/volume/reward_calculator.py
```diff
@@ -0,0 +1,479 @@
...
+
+
+def calc_volume_rewards(
```
trentmc, line 447: We still seem to have two methods to calculate volume rewards:
- "calc_volume_rewards()" -- csvs in, dict out. (Loads csvs then calls RewardCalculator.calculate)
- "RewardCalculator.calculate()" -- dicts in, dict out

Datapoint: dftool_module.py does lots of csv loading & saving. That was the intent. Therefore calc_volume_rewards() violates the intent. 

There are two possible fixes.

Candidate A:
- move {loading of csvs, calc of prev week & cur_week} from calc_volume_rewards() into dftool_module.py
- have dftool_module _directly_ call RewardCalculator.calculate()
- (and therefore deprecated calc_volume_rewards()) directly

Candidate B:
- rename "calc_volume_rewards()" to calc_volume_rewards_from_csvs()
- create a new method "calc_volume_rewards_from_dicts()" which is a one-line pass-through to RewardCalculator.calculate().

I believe Candidate B will play better with the unit tests, because then you can easily test csv-based and non-csv-based flows.

df_py/volume/reward_calculator.py
```diff
@@ -0,0 +1,479 @@
...
+    MAX_N_RANK_ASSETS,
+    PREDICTOOR_MULTIPLIER,
+    RANK_SCALE_OP,
+)
+from df_py.volume import allocations
+from df_py.volume import cleancase as cc
+from df_py.volume import csvs, to_usd
+
+# Weekly Percent Yield needs to be 1.5717%., for max APY of 125%
+TARGET_WPY = 0.015717
```
trentmc, line 25: This should get moved to util.constants.

trentmc (changes requested):

df_py/volume/reward_calculator.py
calina-c: this was done for easier testing, but I managed to find a workaround and I will merge them.

df_py/volume/reward_calculator.py
calina-c: it was like that... I renamed it because I needed some clarity, but I renamed it back :)

df_py/volume/reward_calculator.py
calina-c: I implemented candidate B, and so added back the calc_rewards.py file for conformity with the other modules, containing calc_volume_rewards_from_csvs. However, I don't think calc_volume_rewards_from_dicts is needed, since it would be just a call to RewardCalculator().calculate(). Direct usage of the dict-based calculations is only done in tests, so I think adding an entrypoint like this would not be worth its code.

df_py/volume/calc_rewards.py
```diff
@@ -1,505 +1,54 @@
...
-    for reward_dict in reward_dicts:
-        for key, value in reward_dict.items():
-            merged_dict[key] = merged_dict.get(key, 0) + value
-
-    return merged_dict
-
-
-def calc_rewards_volume(
-    CSV_DIR, START_DATE, TOT_OCEAN, DO_PUBREWARDS=DO_PUBREWARDS, DO_RANK=DO_RANK
+def calc_volume_rewards_from_csvs(
```
trentmc, line 13: This function in line with what we discussed, great.

I just realized: it makes sense to make it csv-in _and_ csv_out.

If you look at dftool_module.py, right after this function is called, it calls:
```python
        csvs.save_volume_rewards_csv(rewperlp, csv_dir)
        csvs.save_volume_rewardsinfo_csv(rewinfo, csv_dir)
```

To me, it makes sense to put those into the function too.

----

And then: make this function call a non-csv version of the function: `calc_volume_rewards()` which is a super-thin wrapper of the OO implementation. I'd put `calc_volume_rewards()` in this module as well (calc_rewards.py).

Then the functions' roles and their naming is obvious:
- calc_volume_rewards_from_csvs
- calc_volume_rewards

df_py/volume/reward_calculator.py
```diff
@@ -0,0 +1,438 @@
...
+                    LP_addrs.add(LP_addr)
+
+        return sorted(LP_addrs)
+
+    @freeze_attributes
+    @enforce_types
+    def _get_predictoors(self) -> Dict[int, List[str]]:
+        """
+        @return
+          predictoors -- dict of (chainID, list of nft_addrs)
```
trentmc, line 363: First: the name of this method is misleading. "Predictoors" are the people making predictions, or the bots that they're running. This method isn't returning predictoors. It's returning the contracts for the prediction feeds. 

To fix, while being aligned with query.py naming, I recommend renaming this method to: `_get_predictoor_contracts()`.

----
Next: It has super-important filtering behavior that's not obvious from its description or implementation.

I recommend to add to the description, something like:

```python
"""
@notes 
  This will only return the prediction feeds that are owned by DEPLOYER_ADDRS, due to functionality of query_predictoor_contracts().
"""
```

----
And, I see that `queries.py::query_predictoor_contracts()` doesn't have a similar note in its description. It should. Please add one.

----

Finally, I couldn't find a unit test for this method. It should have one. It could be simple, by mocking the call to `queries.py::query_predictoor_contracts()`.  

----
One more thing: are there other new functions or methods without unit tests? If yes, let's write basic tests for them too.

trentmc (changes requested):

df_py/volume/reward_calculator.py
calina-c: @trentmc the filtering does not only return feeds owned by deployer addrs. It only uses the keys of deployer addrs, which is the list of chain ids where predictoor contracts are deployed. It is useless and error-prone to query contracts on chains where they are not deployed. I just added this to clarify.

In regards to unit tests, I think only this new function requires testing, otherwise the behaviour is tested (previous tests were adapted).

trentmc: Um, I'm pretty sure it _does_ filter. (If it doesn't, it's a bug.)

From `queries.py::query_predictoor_contracts()`:
```python
        for contract in predictoor_contracts:
            owner = contract["token"]["nft"]["owner"]["id"]
            if chain_id != DEV_CHAINID and owner not in DEPLOYER_ADDRS[chain_id]:
                continue
```

df_py/volume/reward_calculator.py
trentmc: OK

df_py/volume/calc_rewards.py
```diff
@@ -1,505 +1,92 @@
...
+    )
 
-    for reward_dict in reward_dicts:
-        for key, value in reward_dict.items():
-            merged_dict[key] = merged_dict.get(key, 0) + value
+    csvs.save_volume_rewards_csv(rewperlp, str(csv_dir))
+    csvs.save_volume_rewardsinfo_csv(rewinfo, str(csv_dir))
 
-    return merged_dict
+    return rewperlp, rewinfo
```
trentmc, line 57: This should not return anything in memory, because its "return" value is the csvs. Please remove.

Csv in, csv out, that's all.

df_py/volume/calc_rewards.py
trentmc: It's great to see these two functions side-by-side.

Now that I see them, they have some duplicate logic on handling defaults. DRY violation. And it's not consistent, only some logic is duplicated, not all.

I recommend: make the csvs function just a pass-through to the non-csvs function. All that csvs should do is:
- load csvs
- call non-csv function
- save csvs. 

Then have the non-csvs function handle the defaults. Basically, all the functionality that was in the csvs function should be in the non-csvs function:
```python
    prev_week = 0
    if start_date is None:
        cur_week = get_df_week_number(datetime.now())
        prev_week = cur_week - 1
    else:
        prev_week = get_df_week_number(start_date)

    if tot_ocean is None:
        tot_ocean = 0.0

    if do_pubrewards is None:
        do_pubrewards = DO_PUBREWARDS

    if do_rank is None:
        do_rank = DO_RANK
```

trentmc (changes requested):
Very close! 5 min of changes left, probably. :)

df_py/volume/reward_calculator.py
```diff
@@ -0,0 +1,443 @@
...
+
+        @notes
+          The keys in DEPLOYER_ADDRS are the chain ids
+          where predictoor contracts are deployed.
+          There is no point in query-ing on other chains
+        """
+        chainIDs = list(self.stakes.keys())
+        predictoors: Dict[int, List[str]] = {chain_id: [] for chain_id in chainIDs}
```
trentmc, line 371: Recall "predictoors" and "prediction contracts" are very different things. Predictoors are the people making predictions (or their bots). Predictoor contracts are the ERC20 datatoken contracts for prediction feeds.

You've already renamed from `get_predictoors()` to `get_predictoor_contracts()`. Great!

I guess I could have been more explicit about the other changes needed. They include:
- In `get_predictoor_contracts()` code: variable `predictoors` -> `predictoor_contracts`
- In `get_predictoor_contracts()` docstring: variable `predictoors` -> `predictoor_contracts`
- Attribute `self.predictoors` -> `self.predictoor_contracts`
- More? Please check

trentmc (changes requested):

trentmc (approved):

PR #777, opened 2023-11-30
Replace ganache hardfork var name.

calina-c:


calina-c:
var name will be replaced again in barge itself.

PR #792, opened 2024-01-17
Handle exception when getting block timestamp

### LINKED ISSUE #791: Handle block not found errors in timestamp_to_block
**Describe the bug**
Sapphire RPC will throw block not found for block numbers that are too far in the past.

**To Reproduce**
Run volsym for sapphire network, example: https://github.com/oceanprotocol/df-py/actions/runs/7555123620/job/20569794417?pr=790

**Expected behavior**
Handle the error and assume the timestamp at that block is 0

trizin:
Fixes #791

Changes proposed in this PR:

- Handle exception when getting block timestamp

### DIFF
--- a/df_py/util/test/test_blocktime_eth_mainnet.py
+++ b/df_py/util/test/test_blocktime_eth_mainnet.py
@@ -37,10 +37,10 @@ def test_eth_timestamp_to_block_error_handling(monkeypatch):
     web3 = get_web3(get_rpc_url("mainnet"))
     _get_block = web3.eth.get_block
 
-    def random_error_get_block(*args, **kwargs):
-        if random.randint(1, 6) == 1:
+    def random_error_get_block(number):
+        if random.randint(1, 6) == 1 and number != "latest":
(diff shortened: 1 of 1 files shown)


df_py/util/blocktime.py
```diff
@@ -116,7 +116,11 @@ def __init__(self, target_timestamp):
             self.target_timestamp = target_timestamp
 
         def timeSinceTimestamp(self, block_i):
-            block_timestamp = web3.eth.get_block(int(block_i)).timestamp
+            try:
+                block_timestamp = web3.eth.get_block(int(block_i)).timestamp
+            except Exception as e:
+                print(f"An exception occurred while getting block {block_i}, {e}")
+                block_timestamp = 0
```
trentmc, line 123: So if the exception happens, then `block_timestamp = 0`

And it returns `block_timestamp - self.target_timestamp` = `0 - self.target_timestamp` = `-self.target_timestamp`.

This seems wrong.

trentmc (changes requested):

PR #793, opened 2024-01-18
Resolve block not found issue - Sapphire

trizin:
On January 18th, the active flow [failed](https://github.com/oceanprotocol/df-py/actions/runs/7563356604/attempts/2) due to block not found issue originated from Oasis RPC. This PR: 
- Fixes the block not found issue by setting `block_timestamp = 0` thus returning a negative value from `timeSinceTimestamp` so that f(x) != 0
- Verifies that the identified block number is within a 15-minute range of the target timestamp. Throws an error otherwise.

### DIFF
--- a/df_py/util/test/test_blocktime_eth_mainnet.py
+++ b/df_py/util/test/test_blocktime_eth_mainnet.py
@@ -39,7 +39,7 @@ def test_eth_timestamp_to_block_error_handling(monkeypatch):
 
     current_block = web3.eth.get_block("latest")
     current_block_number = current_block.number
-    blocks_ago = web3.eth.get_block(current_block - 5000)
+    blocks_ago = web3.eth.get_block(current_block_number - 5000)
 
     ts = blocks_ago.timestamp
     block = blocks_ago.number


df_py/util/blocktime.py
```diff
@@ -148,7 +148,29 @@ def timeSinceTimestamp(self, block_i):
...
+            "timestamp_to_block() is returning a block that is too far away"
+        )
+    print(
+        "Returning block number",
+        block_i,
+        "for timestamp",
+        timestamp,
+        "diff",
+        abs(block_timestamp - timestamp),
+    )
```
trentmc, line 173: Does a unit test capture this corner case?

trentmc (changes requested):

df_py/util/blocktime.py
```diff
@@ -120,7 +120,7 @@ def timeSinceTimestamp(self, block_i):
                 block_timestamp = web3.eth.get_block(int(block_i)).timestamp
             except Exception as e:
                 print(f"An exception occurred while getting block {block_i}, {e}")
-                return 0
+                block_timestamp = 0
             return block_timestamp - self.target_timestamp
```
trentmc, line 124: I see now that it's ok for this to return `-self.target_timestamp` in the case of an Exception, because it's handled below.

trentmc (approved):

PR #820, opened 2024-03-21
Replace Goerli Exp Vesting Wallet with Sepolia

### LINKED ISSUE #819: Goerli connection fails the reward flow, remove Goerli
Goerli connection fails the reward flow, remove Goerli

Goerli is used to read the vested values from the vesting contract.

Since Halving Vesting is not started yet, it's safe to remove until we deploy on another chain.

trizin:
Fixes #819

Changes proposed in this PR:

- Replace Goerli Exp Vesting Wallet with a new one on Sepolia


trentmc (changes requested):
Don't disable. Point to a different testnet.

trentmc (approved):

PR #824, opened 2024-03-22
Predictoor 5X Boost

### LINKED ISSUE #822: Implement 5x Predictoor Boost for df-py
### Context

Parent issue: [df-private#93](https://github.com/oceanprotocol/df-private/issues/93) "Roll out 5x Predictoor Boost "

Blog post: https://blog.oceanprotocol.com/introducing-5x-predictoor-boost-in-volume-data-farming-c9a93b33b928

### TODOs

Update the df-py payout code to
- [ ] apply the 5x via dcv_multiplier=1.005 (100.5)%
- [ ] only applying to the first 37.5K OCEAN rewards given out on Volume DF (for the remainder it follows the usual 1x)

trizin:
Fixes #822

Changes proposed in this PR:

- Predictoor 5X Boost
- 
-

### DIFF
--- a/df_py/volume/test/test_calc_rewards.py
+++ b/df_py/volume/test/test_calc_rewards.py
@@ -814,7 +814,7 @@ def _test_volume_reward_calculator_pdr_mult(tmp_path, DF_week, expected_rewards)
         "df_py.volume.reward_calculator.DEPLOYER_ADDRS",
         {chain_id: ""},
     ), patch(
-        "df_py.volume.calc_rewards.get_df_week_number", return_value=df_week
+        "df_py.volume.calc_rewards.get_df_week_number", return_value=DF_week
     ), patch(
         "df_py.volume.calc_rewards.wait_to_latest_block"
     ), patch(


df_py/volume/reward_calculator.py
```diff
@@ -176,6 +182,20 @@ def calc_rewards__usd(self) -> np.ndarray:
...
+            # Total boost is limited by PREDICTOOR_OCEAN_BUDGET
+            # Get the number of predictoor assets and calculate the limit per asset
+            n_predictoor_assets = len(
+                self.predictoor_feed_addrs[SAPPHIRE_MAINNET_CHAINID]
+            )
+            if n_predictoor_assets != 0:
+                OCEAN_boost_limit_predictoor = (
+                    PREDICTOOR_OCEAN_BUDGET / n_predictoor_assets
+                )
+
```
trentmc, line 198: You don't need any of the changes in reward_calculator.py.

Also, you need predictoor multiplier to be 0.201 for weeks before DF82.

Simplest is the following.

1. Remove `PREDICTOOR_MULTIPLIER` from constants.py, so you don't accidentally use it

2. Within `dcv_multiplier.py::calc_dcv_multiplier()`, change

From this:
```python
    if is_predictoor:
        return PREDICTOOR_MULTIPLIER
```

To this:
```python
    if is_predictoor:
        if DF_week < 82:
            return 0.201
        return 1.005
```

That's it! 

(And make sure you have a unit test for DF_week=81 and DF_week=82)

df_py/volume/reward_calculator.py
```diff
@@ -193,10 +213,28 @@ def calc_rewards__usd(self) -> np.ndarray:
...
+                    # amount of dcv boosted
+                    boosted_dcv = min(OCEAN_boost_limit_predictoor, DCV_OCEAN_j)
+                    # amount of dcv remaining
+                    remaining_dcv = max(0, DCV_OCEAN_j - OCEAN_boost_limit_predictoor)
+                    boosted_reward_bound = boosted_dcv * multiplier * 5  # 5X BOOST
+                    remaining_reward_bound = remaining_dcv * multiplier  # Remaining
+                    dcv_bound = (
+                        boosted_reward_bound + remaining_reward_bound
+                    ) * perc_at_ij
+
```
trentmc, line 233: (Like mentioned above, remove all these changes.)

df_py/volume/test/test_calc_rewards.py
```diff
@@ -752,3 +756,90 @@ def mock_multipliers(DF_week, is_predictoor):  # pylint: disable=unused-argument
         assert rewards_info[C1][NA][LP1] == approx(60.3, abs=1e-5)
         assert rewards_info[C2][NB][LP2] == 300
         assert rewards_info[C2][NB][LP3] == 300
+
+
+@enforce_types
+def test_volume_reward_calculator_pdr_boost(tmp_path):
```
trentmc, line 762: Change this test such that it tests both DF_week=81 and DF_week=82. See comment above.

How I'd do it: 
```python
@enforce_types
def test_volume_reward_calculator_pdr_boost_week81(tmp_path):
    _test_volume_reward_calculator_pdr_boost(tmp_path, DF_week=81):

@enforce_types
def test_volume_reward_calculator_pdr_boost_week82(tmp_path):
    _test_volume_reward_calculator_pdr_boost(tmp_path, DF_week=82):

@enforce_types
def _test_volume_reward_calculator_pdr_boost(tmp_path):
    ...
    # (like your implementation, with diff't expected numbers for week 81 vs 82)
```

(I prefer two test functions instead of parameterizing a single test function, when there's only a couple cases. Easier to work with.)

trentmc (changes requested):
I proposed a simplifying change. Sorry for the delay. 

@trizin are you able to change in time?

If urgent, we can merge this and then put my proposal in afterwards.

df_py/volume/reward_calculator.py
trentmc: No decision has changed. You are correct, it only applies to Predictoor.

And the proposed code supports this: that is why it has "if predictoor". 

There is still the non-predictoor section of code below the "if Predictoor" section. It doesn't change.

df_py/volume/test/test_calc_rewards.py
```diff
@@ -752,3 +756,69 @@ def mock_multipliers(DF_week, is_predictoor):  # pylint: disable=unused-argument
         assert rewards_info[C1][NA][LP1] == approx(60.3, abs=1e-5)
         assert rewards_info[C2][NB][LP2] == 300
         assert rewards_info[C2][NB][LP3] == 300
+
+
+@enforce_types
+@pytest.mark.parametrize(
+    "df_week, expected_rewards",
+    [
+        (80, 100.5),
```
trentmc, line 765: Why is it 100.5 and then 100.5 * 5,

Rather than 20.1 and then 20.1 * 5?

